PR #122, opened 2017-04-12
clarity improvements for README

drh-stanford:


README.md
```diff
@@ -28,25 +32,32 @@ TODO: Write usage instructions here
...
 
 ```sh
 $ bundle exec rake arclight:server
 ```
 
+Then visit http://localhost:3000.
+
 ### Run the test suite
 
+Ensure Jetty and Rails are not running, then:
```
mejackreed: This now uses solr_wrapper instead of jetty. Perhaps we can switch out jetty for something else here?

drh-stanford: How about?:

### Run a development server

```sh
$ bundle exec rake arclight:server
```

Then visit http://localhost:3000. It will also start a Solr instance on port 8983.

### Run the test suite

Ensure Solr and Rails are not running (ports 8983 and 3000 respectively), then:

```sh
$ bundle exec rake
```

mejackreed: 👍

drh-stanford:
Ok I force pushed up those changes.

PR #132, opened 2017-04-14
add home page and tests

### LINKED ISSUE #1: Landing/home page
There should be a landing page for the application, with a prominent search bar. [Additional features for home page to come]

Acceptance criteria:
* I see a landing page and can start a search.
* The landing page does not look like the default Blacklight home page.

gordonleacock:
closes #1 
screen shot 2017-04-14 at 2 39 17 pm

### DIFF
--- a/app/views/catalog/_home.html.erb
+++ b/app/views/catalog/_home.html.erb
@@ -1,0 +1,8 @@
+<div class='jumbotron'>
+  <div class='row'>
+    <h1 class='text-center'><%= t('arclight.views.home.heading') %></h1>
+  </div>
+  <div class='col-md-6 col-md-offset-3'>
+      <%= render_search_bar %>
+  </div>
+</div>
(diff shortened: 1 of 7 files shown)


app/views/catalog/_home.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class='jumbotron'>
+  <div class='row'>
+    <h1><%= t('arclight.views.home.heading') %></h1>
```
jkeck: What would you think about adding `text-center` class here to center this heading?

jkeck:
Looks good. 👌 

One minor comment that you may have already thought about.

app/views/catalog/_home.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  </div>
+  <div class='col-md-6 col-md-offset-3'>
```
jkeck: I wonder if we should make this a bit larger on desktop (just using the bootstrap classes).  Perhaps that got more complicated than I'm making it sound, so totally understand if we want to punt and work on the proper sizing later.

gordonleacock: I'll see if I can add those modifications next.

PR #134, opened 2017-04-14
Add a .solr_wrapper configuration file.

jkeck:
This allows us to start the solr index and the rails app separately while we're developing.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,7 +17,6 @@ Metrics/LineLength:
   Max: 120
   Exclude:
     - 'Gemfile'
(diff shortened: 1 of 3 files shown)


mejackreed:
Did this have any effect slimming down how the rake tasks are configured? 

https://github.com/sul-dlss/arclight/blob/master/tasks/arclight.rake#L13
https://github.com/sul-dlss/arclight/blob/master/tasks/arclight.rake#L39

jkeck:
Ah, yeah, I forgot to check those out.  Will try that out right now.

mejackreed:
So pretty now 🌈 💡

PR #135, opened 2017-04-14
Add Creator and Names facets

### LINKED ISSUE #5: Facet and refine search results by names
* Includes creators and names included as controlled access terms
* Includes personal and corporate names in a single "Names" facets

Ref: req 1.4 ("Facet and refine search results by agent/creator")

drh-stanford:
This PR fixes #5.

screen shot 2017-04-14 at 1 27 08 pm

Note that none of the custom indexing code has specs ... we're testing the display of the fields, but that doesn't work for the facets. should we introduce specs for the custom indexing directly?

### DIFF
--- a/lib/arclight/custom_document.rb
+++ b/lib/arclight/custom_document.rb
@@ -17,5 +17,17 @@ module Arclight
       t.accessrestrict(path: 'archdesc/accessrestrict/p', index_as: %i[displayable])
       t.scopecontent(path: 'archdesc/scopecontent/p', index_as: %i[displayable])
     end
+
+    def to_solr(solr_doc = {})
+      super
+      Solrizer.insert_field(solr_doc, 'names', names, :facetable)
+      solr_doc
+    end
+
+    private
+
+    def names
+      [corpname, famname, name, persname].flatten.compact.uniq
+    end
   end
 end
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -71,6 +71,7 @@ class CatalogController < ApplicationController
 
     config.add_facet_field 'creator_sim', label: 'Creator'
     config.add_facet_field 'level_sim', label: 'Level'
+    config.add_facet_field 'names_sim', label: 'Names'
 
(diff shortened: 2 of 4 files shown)


jkeck:
@drh-stanford I believe we should be able to write tests for the facet values.  We may just need to do something like `visible: false` in our tests (although I thought Blacklight would spit the facet values onto the page when JS was not available).

lib/arclight/custom_document.rb
```diff
@@ -17,5 +17,17 @@ class CustomDocument < SolrEad::Document
       t.accessrestrict(path: 'archdesc/accessrestrict/p', index_as: %i[displayable])
       t.scopecontent(path: 'archdesc/scopecontent/p', index_as: %i[displayable])
     end
+
+    def to_solr(solr_doc = {})
+      super(solr_doc)
+      Solrizer.insert_field(solr_doc, 'names', names, :facetable)
```
mejackreed: Gosh it feels sad to me that there is no way in OM to combine fields. 😢

drh-stanford: We could write a long XPath expression that would do the combination. Would that be an improvement?

mejackreed: Probably not. I like what you did, just which there was an api to do it cleaner.

drh-stanford:
ok i force pushed up some revised specs that test for facet values.

lib/arclight/custom_document.rb
```diff
@@ -17,5 +17,17 @@ class CustomDocument < SolrEad::Document
       t.accessrestrict(path: 'archdesc/accessrestrict/p', index_as: %i[displayable])
       t.scopecontent(path: 'archdesc/scopecontent/p', index_as: %i[displayable])
     end
+
+    def to_solr(solr_doc = {})
+      super(solr_doc)
```
jkeck: This might be just a random ruby fact, but `super` calls the parent method with the arguments passed to the current method, so you can simply call `super` (without arguments) here and get the same result.

drh-stanford: ok, i've replaced that `super(solr_doc)` with `super`.

PR #156, opened 2017-04-17
index page shows breadcrumb for parents

### LINKED ISSUE #4: Display breadcrumb for hierarchy for each item/component in a search result
Each item/component should be aware of how to get information about its parent components and collection. 

Acceptance criteria:
* Each component in a set of search results should display the breadcrumb, and link them to the relevant show views for those components

mejackreed:
Closes #4 

hier

### DIFF
--- a/app/models/arclight/parents.rb
+++ b/app/models/arclight/parents.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Arclight
+  ##
+  # Object for parsing and formalizing Solr_Ead "Parents"
+  # https://github.com/awead/solr_ead/blob/8cf7ffaa66e0e4c9c0b12f5646d6c2e20984cd99/lib/solr_ead/behaviors.rb#L54-L57
+  class Parents
+    attr_reader :ids, :labels, :eadid
+    def initialize(ids:, labels:, eadid:)
+      @ids = ids
+      @labels = labels
+      @eadid = eadid
+    end
+
+    ##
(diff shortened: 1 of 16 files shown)


app/models/arclight/parents.rb
```diff
@@ -0,0 +1,30 @@
...
+    def as_parents
+      Hash[ids.zip(labels)].map { |k, v| Arclight::Parent.new(id: k, label: v, eadid: eadid) }
+    end
+
+    ##
+    # @param [SolrDocument] document
+    def self.from_solr_document(document)
+      ids = document.fetch('parent_ssm', [])
+      labels = document.fetch('parent_unittitles_ssm', [])
+      eadid = document.fetch('ead_ssi', '')
```
jkeck: Any thoughts on making these fields configurable in some way?  (or not hard-coded in the method)

mejackreed: Yeah I definitely think we should, but we need a global approach. Right now this is happening in multiple places:

 - magically in solr_ead/om
 - `solrconfig.xml`
 -  generated catalog controller
 -  our custom indexing code
 -  this file

One approach here would to drop some behavior into `SolrDocument` to start, but that doesn't solve our problems everywhere else.

jkeck: Yeah, and I'm not sure that we'll be able to get an approach that accounts for all the places outside of our control, but not hard-coding solr fields in our methods could be a good start.

PR #157, opened 2017-04-17
add repository to facet and remove it from bundled Names facet

### LINKED ISSUE #12: Facet and refine search results by repository
Search results should facetable by repository.

**Acceptance criteria**

* Clicking a specific facet value in the "Repository" facet limits results to collections/components matching that facet value.

- - - 
* Ref: req 1.9
* Category: 1. Core Discovery
* Subcategory: Browse

### LINKED ISSUE #140: Repository should not appear in the "Names" facet
Currently, the indexing rules seem to be matching `<corpname/>` elements inside the `<repository/>` tag. They shouldn't; repository should be a distinct facet (see #12).

drh-stanford:
This PR fixes #12 and fixes #140.

Note that it requires https://github.com/awead/solr_ead/pull/10 for the facet to work correctly.

screen shot 2017-04-17 at 2 52 33 pm

### DIFF
--- a/spec/features/arclight_spec.rb
+++ b/spec/features/arclight_spec.rb
@@ -57,6 +57,8 @@ RSpec.describe 'Arclight', type: :feature do
         expect(page).to have_css('h3 a', text: 'Creator')
         expect(page).to have_css('li .facet-label', text: 'Alpha Omega Alpha', visible: false)
         expect(page).to have_css('h3 a', text: 'Names')
+        expect(page).to have_css('li .facet-label', text: 'Root, William Webster, 1867-1932', visible: false)
+        expect(page).to have_css('h3 a', text: 'Repository')
         expect(page).to have_css('li .facet-label', text: '1118 Badger Vine Special Collections', visible: false)
       end
     end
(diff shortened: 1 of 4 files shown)


spec/features/arclight_spec.rb
```diff
@@ -55,6 +55,8 @@
         expect(page).to have_css('h3 a', text: 'Creator')
         expect(page).to have_css('li .facet-label', text: 'Alpha Omega Alpha', visible: false)
         expect(page).to have_css('h3 a', text: 'Names')
+        expect(page).to have_css('li .facet-label', text: 'Root, William Webster, 1867-1932', visible: false)
```
jkeck: Perhaps we can use the class applied to the panel (which infers the field name) to constrain these tests to ensure that we've met the acceptance criteria.

As the test is written, it would match if the facet value existed in the original facet (which the acceptance criterial explicitly calls out).

drh-stanford:
Ok, I rebased and force pushed a change to the spec such that all facets are constrainted to their specific panel. e.g.,

```ruby
      within('#facets') do
        within('.blacklight-repository_sim') do
          ... expects ...
        end
     end
```

jkeck:
WRT to the Rubocop failure, I wonder if we should exclude this file from the line-length cop.  xpath queries can get long, and I'm not sure we'll be doing ourselves any favors by trying to limit line length.

drh-stanford:
Sorry, I forgot to run rubocop ☹️ I added an exclusion for the indexing code.

Issue #142, opened 2017-04-17
Facet and refine search results by collection


anarchivist:
Facet and refine search results by the collection title, for both collections and components

jkeck:
Closed by #158

PR #158, opened 2017-04-17
add collection facet

gordonleacock:
It was a bit hard to tell if this meets all the requirements of issue [#142](https://github.com/sul-dlss/arclight/issues/142) because there is only one collection.
Also, although it appeared to get all the components in collection I wasn't sure how to double-check that.
Finally, I commented out the spec I added because it seemed close but **didn't work.**

I'm glad to make any changes needed.


drh-stanford:
We've been adding the new facet fields to the default Solr query handler -- see here https://github.com/sul-dlss/arclight/blob/master/solr/conf/solrconfig.xml#L209-L211

gordonleacock:
Thanks Darren.

PR #159, opened 2017-04-17
add the Level facet with machine/human-readable values

### LINKED ISSUE #14: Facet and refine search results by level of description
**Acceptance criteria**

* Collection-level facet value should be "Collection"
* Component-level facet value should be derived from `level` attribute
    * If `level="otherlevel"`, use the value from the `otherlevel` attribute. if `otherlevel` is missing, use `Other`
    * Values should be capitalized

- - - 

* Ref: req 1.62
* Category: 1. Core Discovery
* Subcategory: Browse

drh-stanford:
This PR fixes #14. 

Note that `level_ssm` is as-is data from `@level` and `level_sim` is the human readable version. It also adds some new specs for our custom indexing code. 

Note that the CustomComponent spec is a little hairy due to how `solr_ead` works for components. We can refactor it to look more like the CustomDocument code, but AFAIK we'd have to create fixture files that just had the XML for the component.

## What it looks like:

screen shot 2017-04-17 at 2 45 20 pm

"Mixed File" is the special case when level=otherlevel and otherlevel is present -- see the XML fixture change

### DIFF
--- a/spec/features/indexing_custom_component_spec.rb
+++ b/spec/features/indexing_custom_component_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Indexing Custom Component', type: :feature do
+  let(:components) do # an Array<Arclight::CustomComponent>
+    options = { component: Arclight::CustomComponent }
+    ENV['SOLR_URL'] = Blacklight.connection_config[:url]
+    indexer = SolrEad::Indexer.new(options) # `initialize` requires a solr connection
+
+    components = []
+    indexer.components('spec/fixtures/ead/alphaomegaalpha.xml').each do |node|
+      components << indexer.send(:om_component_from_node, node) # private method :(
+    end
+    components
+  end
+
+  context 'solrizer' do
+    it '#level' do
+      doc = components[0].to_solr
+      expect(doc['level_ssm'].first).to eq 'series'
+      expect(doc['level_sim'].first).to eq 'Series'
+
+      doc = components[2].to_solr
+      expect(doc['level_ssm'].first).to eq 'otherlevel'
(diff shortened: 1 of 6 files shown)


drh-stanford:
We're not spec'ing our custom indexing code, and that's why coveralls complains. I'll writeup a spec for it...

spec/features/indexing_custom_component_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Indexing Custom Component', type: :feature do
+  let(:components) do # an Array<Arclight::CustomComponent>
+    options = { component: Arclight::CustomComponent }
+    ENV['SOLR_URL'] = 'http://127.0.0.1:8983/solr/blacklight-core'
```
jkeck: The fact that we have this hard-coded in the rake task was curious to me.  Is there a way we can set this w/o hard-coding the solr URL (and perhaps we can use the same pattern in other places).

drh-stanford: Yeah, `solr_ead` requires a Solr connection during its initialization. We could also try to put in a PR that does a lazy eval for the Solr connection. Rather than the `SOLR_URL` env varaible, it will take a `solr.yml`. Not sure which is the better approach.

jkeck: Totally understand.  In the meantime, can we set the `SOLR_URL` to be the value from our configured solr URL? (which I believe is available to us through the global `Blacklight` object)

drh-stanford: Ok, I was able to find it in `Blacklight.connection_config[:url]`. Force pushed an update for that.

jkeck:
Love that 💯 % coverage! 😎

PR #164, opened 2017-04-19
Fixture validation and addition of normalized dates

anarchivist:


anarchivist:
The build should be failing at this point, because the EAD is currently invalid. I'll add a new commit that fixes that issue (and update the associated tests).

.travis.yml
```diff
@@ -7,3 +7,9 @@ rvm:
 jdk:
   - oraclejdk8
 cache: bundler
+before_script:
+  - xmllint --noout --schema spec/fixtures/xsd/ead.xsd spec/fixtures/ead/*.xml
```
jkeck: Does Travis not allow us to lint using the xsd @ loc.gov?

anarchivist: We are able to run `xmllint` as follows from Travis: 

```bash
xmllint --noout --schema http://www.loc.gov/ead/ead.xsd spec/fixtures/ead/*.xml
```

The cached copy of the XSD is really just a precautionary measure. It's been my experience that in cases where the US Government shuts down, XML-based metadata stops validating because they point their web servers somewhere else and the schemas are no longer dereferenceable.

jkeck: Oh but that _would never happen._ :trollface:

PR #167, opened 2017-04-19
Add a date range facet

### LINKED ISSUE #8: Facet and refine search results by dates
Users should be able to facet and refine search results by dates.

**Acceptance criteria**

* The facet should display as a range facet
* The range for the facet query should be controllable by two sliders OR by two text entry widgets
* Fixture data requirement: Must contain normalized dates #151
* NOTE that the "top-level" date histogram is another ticket: #141
- - -

* Ref: req 1.6
* Category: 1. Core Discovery
* Subcategory: Browse

jkeck:
Closes #8 

Outstanding questions:
- [x] Should we introduce a javascript testing framework so we can write a feature spec that the date range code is properly installed and configured in our app?
- [x] Has there been any thought/discussion around reuse of code between our `CustomComponent` and `CustomDocument` classes?

range-limit

### DIFF
--- a/arclight.gemspec
+++ b/arclight.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'capybara'
   spec.add_development_dependency 'coveralls'
   spec.add_development_dependency 'engine_cart'
+  spec.add_development_dependency 'poltergeist'
   spec.add_development_dependency 'rake', '~> 12.0'
   spec.add_development_dependency 'rubocop', '~> 0.48.1'
   spec.add_development_dependency 'rubocop-rspec', '~> 1.15.0'
--- a/arclight.gemspec
+++ b/arclight.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_dependency 'blacklight'
+  spec.add_dependency 'blacklight_range_limit'
   spec.add_dependency 'rails', '>= 4.2.6', '< 6'
   spec.add_dependency 'solr_ead'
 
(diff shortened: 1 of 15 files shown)


mejackreed:
Yes to bring in poltergeist.

drh-stanford:
No, afaik we haven't talked about sharing code between CustomDocument and CustomComponent. They don't subclass off the same class, but we could add a mixin helper module.

jkeck:
If folks would be 🆒 w/ a mixin then I'm more than happy to do that.

drh-stanford:
Also, I noticed that the `template.rb` file got merged so we should probably add the blacklight range limit install and the github pin to it -- see https://github.com/sul-dlss/arclight/blob/master/template.rb. I dunno if we can get engine_cart to use our `template.rb` instead of the test_app_generator.rb so that we don't have that procedure in 2 places.

jkeck:
Ah, good catch.  I'll do that ASAP.

jkeck:
Hmm, actually, it looks like the template calls the `arclight:install` generator which I have added the blacklight_range_limit installer to already: https://github.com/sul-dlss/arclight/pull/167/files#diff-51a9d458b1ab3862a4029a2dd0278d3fR21.

I _think_ this should just work, no?

mejackreed:
This is always a hard question. GeoBlacklight removed range limit in favor of documentation. We started running into a lot of issues where things were not working correctly for us.

jkeck:
I added a reference to `template.rb` to install the `blacklight_range_limit` gem from the appropriate branch.

I also added a comment about potentially not needing to explicitly require the gems in our `template.rb` once versions are released that have the behavior we need.

Issue #144, opened 2017-04-17
Collection detail page - overview with full detail and navigation


anarchivist:
This is a refinement of #138. This adds multiple headings for groupings of additional data, and a navigator to the left-hand side of the page. 

**Acceptance criteria**

* Display of the following section headings, with the corresponding data:
    * Summary (see #138)
    * Access and Use (see #138)
    * Biographical / Historical (same as "Background" in mockup) - data from `<bioghist/>`
    * Scope and Arrangement 
         * Scope and content: `<scopecontent/>`
         * Arrangement: `<arrangement/>`
    * Related
         * Related material: `<relatedmaterial/>`
         * Separated material `<separatedmaterial/>`
         * Other Finding Aids `<otherfindaid/>`
         * Alternative Form Available `<altformavail/>`
         * Location of originals `<originalsloc/>`
    * Indexed terms
         * Specific children of `<controlaccess/>`: `<corpname/>`, `<famname/>`, `<function/>`, `<genreform/>`, `<geogname/>`, `<occupation/>`, `<persname/>`, `<subject/>`, and `<title/>`
    * Administrative Information
         * Acquisition information: `<acqinfo/>`
         * Appraisal information: `<appraisal/>`
         * Custodial history: `<custodhist/>`
         * Processing Information: `<processinfo/>`
* A navigation bar in a lefthand sidebar, with entries for each heading that link to those headings.
* The navigation aid should also exhibit "scrollspy" behavior, highlighting the currently active section.

**Mockups**

![collection-detail-overview-markedup](https://cloud.githubusercontent.com/assets/73732/25079069/f5d3d6f2-22ed-11e7-9bba-857267a6409f.jpg)

![collection-detail-overview-no-inventory-markedup](https://cloud.githubusercontent.com/assets/73732/25079071/fbaf007e-22ed-11e7-95a5-698d6d9ea7cf.jpg)

jkeck:
@anarchivist it looks like our current fixtures does not have any `arrangement` data.  I'm not 100% sure what kind of data should go in there if you want to shoot me something I can add it in my PR so I can write tests that the data shows up.

jkeck:
Also, it appears that none of the data for the `Related` section is available in our current fixture.
`<relatedmaterial/>`, `<separatedmaterial/>`, `<otherfindaid/>`, `<altformavail/>` and `<originalsloc/>` are all not present.

anarchivist:
Working on that now…

[...]

Issue #11, opened 2017-04-10
Facet and refine search results by subject


anarchivist:
**Acceptance criteria**

* Index data from the children of `<controlaccess/>`: `<subject/>`, `<function/>`, `<genreform/>`, `<occupation/>`
* Selecting the facet should filter results by the selected value
* [ ] _Question_: Do these need to be indexed into separate fields?

- - -

* Ref: req 1.11
* Category: 1. Core Discovery
* Subcategory: Browse

anarchivist:
87a81e1 adds data for the specific additional fields.

jkeck:
@anarchivist, we notice that the subject shows up under the collection and the component, but it's not specified here which one we want (or if it is both).

The xpath in the current PR indexes any control access sub-elements into the collection (even when they are elements in a component).

anarchivist:
> The xpath in the current PR indexes any control access sub-elements into the collection (even when they are elements in a component). 

I think this is correct for this ticket.

jkeck:
👌

jkeck:
Closed by #169

anarchivist:
OK, going back and looking at #179, I think this was wrong, but I'll break out the changes into a new ticket.

PR #169, opened 2017-04-19
add multifield facet based on subfields of controlaccess

gordonleacock:
Adopted code from https://github.com/NYULibraries/ead_indexer/blob/master/lib/ead_indexer/behaviors.rb#L55-L58 to add a multi field facet based on subfields of archdesc/controlaccess/

### DIFF
--- a/lib/arclight/custom_document.rb
+++ b/lib/arclight/custom_document.rb
@@ -12,6 +12,8 @@ module Arclight
       t.repository(path: 'archdesc/did/repository/corpname/text() | archdesc/did/repository/name/text()', index_as: %i[displayable facetable])
       t.creator(path: "archdesc/did/origination[@label='creator']/*/text()", index_as: %i[displayable facetable])
       t.prefercite(path: 'archdesc/prefercite/p', index_as: %i[displayable])
+      t.function(path: 'archdesc/controlaccess/function/text()', index_as: %i[displayable facetable])
+      t.occupation(path: 'archdesc/controlaccess/occupation/text()', index_as: %i[displayable facetable])
 
       # overrides of solr_ead to get different `index_as` properties
       t.extent(path: 'archdesc/did/physdesc/extent', index_as: %i[displayable])
@@ -29,6 +31,7 @@ module Arclight
       Solrizer.insert_field(solr_doc, 'level', 'Collection', :facetable) # human-readable
       Solrizer.insert_field(solr_doc, 'names', names, :facetable)
       Solrizer.insert_field(solr_doc, 'date_range', formatted_unitdate_for_range, :facetable)
+      Solrizer.insert_field(solr_doc, 'access_subjects', access_subjects, :facetable)
       solr_doc
     end
 
@@ -36,6 +39,35 @@ module Arclight
 
     def names
       [corpname, famname, name, persname].flatten.compact.uniq - repository
+    end
+
+    # Combine subjets into one group from:
+    #  <controlaccess/><subject></subject>
+    #  <controlaccess/><function></function>
+    #  <controlaccess/><genreform></genreform>
+    #  <controlaccess/><occupation></occupation>
+    def access_subjects
+      subjects = search("//*[local-name()='subject' or local-name()='function' or local-name() = 'occupation' or local-name() = 'genreform']").to_a
+      clean_facets_array(subjects.flatten.map(&:text))
+    end
+
+    # Return a cleaned array of facets without marc subfields
+    #
+    # E.g. clean_facets_array(['FacetValue1 |z FacetValue2','FacetValue3']) => ['FacetValue1 -- FacetValue2', 'FacetValue3']
+    def clean_facets_array(facets_array)
+      Array(facets_array).map { |text| fix_subfield_demarcators(text) }.compact.uniq
+    end
+
+    # Replace MARC style subfield demarcators
+    #
+    # Usage: fix_subfield_demarcators("Subject 1 |z Sub-Subject 2") => "Subject 1 -- Sub-Subject 2"
+    def fix_subfield_demarcators(value)
+      value.gsub(/\|\w{1}/, '--')
+    end
+
+    # Wrap OM's find_by_xpath for convenience
+    def search(path)
+      find_by_xpath(path) # rubocop:disable DynamicFindBy
     end
   end
 end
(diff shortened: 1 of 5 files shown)


gordonleacock:
Okay, now fixing some rubo cop issues!

spec/features/arclight_spec.rb
```diff
@@ -84,6 +84,11 @@
           expect(page).to have_css('h3 a', text: 'Place')
           expect(page).to have_css('li .facet-label', text: 'Mindanao Island (Philippines)', visible: false)
         end
+
+        within('.blacklight-access_subjects_sim') do
+          expect(page).to have_css('h3 a', text: 'Subject')
+          expect(page).to have_css('li .facet-label', text: 'Fraternizing', visible: false)
```
drh-stanford: Can you add the other data in the fixture to ensure we test that long xpath query in the `access_subjects` method to catch all 4 of those elements? This is the data I see in the fixture right now:

```xml
      <subject source="mesh">Societies</subject>
      <function source="local">Fraternizing</function>
      <genreform source="aat">Photographs</genreform>
      <occupation source="mesh">Medicine</occupation>
```

gordonleacock: Hi Darren - I'm confused because I think all four are there. Those are the only ones issue #11 specifies. Am I missing your point?
<controlaccess>
      <subject source="mesh">Societies</subject>
      <function source="local">Fraternizing</function>
      <genreform source="aat">Photographs</genreform>
      <occupation source="mesh">Medicine</occupation>
      <corpname role="fmo" source="ingest">Alpha Omega Alpha</corpname>
      <persname source="mesh">Root, William Webster, 1867-1932</persname>
      <persname source="mesh">Bierring, Walter L. (Walter Lawrence), 1868-1961</persname>
      <geogname source="lcsh">Mindanao Island (Philippines)</geogname>
    </controlaccess>

drh-stanford: Sorry that I wasn't clear. I'm saying to add the other access_subjects values to the spec so that we can verify the indexing code is picking them out of all those elements, e.g.,

```ruby
expect(page).to have_css('li .facet-label', text: 'Societies', visible: false) # from <subject/>
expect(page).to have_css('li .facet-label', text: 'Fraternizing', visible: false) # from <function/>
expect(page).to have_css('li .facet-label', text: 'Photographs', visible: false) # from <genreform/>
expect(page).to have_css('li .facet-label', text: 'Medicine', visible: false) # from <occupation/>
```

lib/arclight/custom_document.rb
```diff
@@ -31,5 +34,38 @@ def to_solr(solr_doc = {})
...
+    def fix_subfield_demarcators(value)
+      value.gsub(/\|\w{1}/, '--')
+    end
+
+    # Wrap OM's find_by_xpath for convenience
+    def search(path)
+      find_by_xpath(path) # rubocop:disable DynamicFindBy
+    end
+
+    def value(path)
```
jkeck: Is this method being used anywhere? (not sure if this is an override of a parent class or a mixin or not).

This method is not being executed in the context of our test suite which makes me wonder if it's superfluous.

spec/features/arclight_spec.rb
gordonleacock: Good idea @drh-stanford - I've added those!

lib/arclight/custom_document.rb
```diff
@@ -10,6 +10,8 @@ class CustomDocument < SolrEad::Document
       t.unitid(path: 'archdesc/did/unitid', index_as: %i[displayable])
       t.repository(path: 'archdesc/did/repository/corpname/text() | archdesc/did/repository/name/text()', index_as: %i[displayable facetable])
       t.creator(path: "archdesc/did/origination[@label='creator']/*/text()", index_as: %i[displayable facetable])
+      t.function(path: 'archdesc/controlaccess/function/text()', index_as: %i[displayable facetable])
+      t.occupation(path: 'archdesc/controlaccess/occupation/text()', index_as: %i[displayable facetable])
```
jkeck: Are `function` or `occupation` solr fields that we're generating here being used?

I see that we're also parsing them directly via xpath in the `access_subjects` method below.  Sorry if this is just my OM/Solrizer naivety on how these things work.

lib/arclight/custom_document.rb
```diff
@@ -31,5 +34,34 @@ def to_solr(solr_doc = {})
     def names
       [corpname, famname, name, persname].flatten.compact.uniq - repository
     end
+
+    # Combine subjets into one group from:
+    #  <controlaccess/><subject></subject>
+    #  <controlaccess/><function></function>
+    #  <controlaccess/><genreform></genreform>
+    #  <controlaccess/><occupation></occupation>
+    def access_subjects(subjects = [])
```
jkeck: Is there a case where we pass in an array of subjects to the `access_subjects` method?  (I couldn't find any looking through the code briefly)

gordonleacock: @jkeck - Not that I know of. I was followed the NYU Library code: https://github.com/NYULibraries/ead_indexer/blob/master/lib/ead_indexer/behaviors.rb#L55-L58 which initializes the subject variable as a param to the def.

Would you like me to change that?

jkeck: Something that might be a bit more idiomatic would be to not have the method take any arguments and assign a local `subjects` variable in the method to be the results of the `search` function coerced into an array (using `to_a`).

That would make the like 44 look something like
```ruby
subjects = search("//*[local-name()...").to_a
```

lib/arclight/custom_document.rb
jkeck: Ah, I actually see that we'll need them for #144 anyway, so 👍 to keeping this in here.

lib/arclight/custom_document.rb
gordonleacock: @jkeck - Yep, that works!

lib/arclight/custom_document.rb
```diff
@@ -37,5 +40,34 @@ def to_solr(solr_doc = {})
...
       [corpname, famname, name, persname].flatten.compact.uniq - repository
     end
+
+    # Combine subjets into one group from:
+    #  <controlaccess/><subject></subject>
+    #  <controlaccess/><function></function>
+    #  <controlaccess/><genreform></genreform>
+    #  <controlaccess/><occupation></occupation>
+    def access_subjects
+      subjects = search("//*[local-name()='subject' or local-name()='function' or local-name() = 'occupation' or local-name() = 'genreform']").to_a
```
jkeck: This xpath will index controlaccess sub-elements from within components into the collection (e.g. `//*`).  I have a question out to @anarchivist on #11 for clarification.

jkeck: See https://github.com/sul-dlss/arclight/issues/11#issuecomment-295939352

Looks like we're good.

PR #170, opened 2017-04-19
move arclight:index* tasks into shared tasks and some minor cleanup

drh-stanford:
This PR lets the `arclight-demo` (or any other arclight) application use the `arclight:index*` tasks.

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'arclight'
+
+namespace :arclight do
+  desc 'Index a document'
+  task :index do
+    raise 'Please specify your file, ex. FILE=<path/to/file.xml>' unless ENV['FILE']
+    indexer = load_indexer
+    indexer.update(ENV['FILE'])
+  end
+
+  desc 'Index a directory of documents'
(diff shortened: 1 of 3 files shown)


lib/tasks/index.rake
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'arclight'
+
+# hardcoded since we don't have access to Blacklight.connection_config[:url] here
+SOLR_URL = ENV['SOLR_URL'] || 'http://127.0.0.1:8983/solr/blacklight-core'
```
jkeck: I wonder if this would be better written as `ENV['SOLR_URL'] ||= 'http...` and then not set a local `SOLR_URL` constant at all.

If we did that we could kill line 29 in this file as well I think.

jkeck: And maybe that line just lives in the `load_indexer` method below?  Just thinking out-loud.

drh-stanford: Ok, good idea. I changed it to 
```
ENV['SOLR_URL'] ||= 'http://127.0.0.1:8983/solr/blacklight-core'
``` 
in the load_indexer method

PR #176, opened 2017-04-20
add Date to Collection Title

### LINKED ISSUE #172: Collection detail page: The collection's title and dates are displayed prominently in a single heading

drh-stanford:
This PR fixes #172. I think this is the blacklight-way of doing this...

screen shot 2017-04-20 at 12 20 53 pm

### DIFF
--- a/app/presenters/arclight/show_presenter.rb
+++ b/app/presenters/arclight/show_presenter.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Arclight
+  # Custom presentation methods for show partial
+  class ShowPresenter < Blacklight::ShowPresenter
+    def heading
+      title = super
+      date = document.fetch('unitdate_ssm', [])
+      title += ", #{date.first}" if date.present?
+      title
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


app/presenters/arclight/show_presenter.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Arclight
+  # Custom presentation methods for show partial
+  class ShowPresenter < Blacklight::ShowPresenter
+    def heading
+      title = super
+      date = document.fetch('unitdate_ssm', [])
+      title += ", #{date.first}" if date.present?
```
mejackreed: Is the `if data.present?` needed here? Won't `fetch` return `[]` if its not there?

drh-stanford: Not sure I understand. I don't want to append a `, ` if there's no date -- that's what that `if` clause is trying to do.

mejackreed: Gotcha.. sorry I missed that :)

PR #177, opened 2017-04-20
Add additional metadata to the collection show page

### LINKED ISSUE #144: Collection detail page - overview with full detail and navigation

jkeck:
Closes #144 (breaking remaining tasks into separate issues #178 & #179)

collection-full-details

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -141,7 +141,13 @@ class CatalogController < ApplicationController
     # Configuration for partials
     config.index.partials.insert(0, :index_breadcrumb)
 
-    config.show.metadata_partials = [:summary_field, :access_field, :background_field, :scope_and_arrangement_field]
+    config.show.metadata_partials = [
+      :summary_field,
+      :access_field,
+      :background_field,
+      :scope_and_arrangement_field,
+      :related_field
+    ]
 
     # Collection Show Page - Summary Section
     config.add_summary_field 'creator_ssm', label: 'Creator'
@@ -160,5 +166,12 @@ class CatalogController < ApplicationController
     # Collection Show Page - Scope and Arrangement Section
     config.add_scope_and_arrangement_field 'scopecontent_ssm', label: 'Scope and Content'
     config.add_scope_and_arrangement_field 'arrangement_ssm', label: 'Arrangement'
+
+    # Collection Show Page - Related Section
+    config.add_related_field 'relatedmaterial_ssm', label: 'Related material'
+    config.add_related_field 'separatedmaterial_ssm', label: 'Separated material'
+    config.add_related_field 'otherfindaid_ssm', label: 'Other finding aids'
+    config.add_related_field 'altformavail_ssm', label: 'Alternative form available'
+    config.add_related_field 'originalsloc_ssm', label: 'Location of originals'
   end
 end
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -146,7 +146,8 @@ class CatalogController < ApplicationController
       :access_field,
       :background_field,
       :scope_and_arrangement_field,
-      :related_field
+      :related_field,
+      :admin_info_field
     ]
 
     # Collection Show Page - Summary Section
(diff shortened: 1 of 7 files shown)


spec/fixtures/ead/alphaomegaalpha.xml
```diff
@@ -83,11 +83,6 @@
         </defitem>
       </list>
     </odd>
-    <acqinfo id="aspace_4179cc95d0af3f4bc52ca7f563e45888">
-      <head>Provenance</head>
-      <p>Gift,John Z. Bowers in 1973; Gladys Brill Brampton in 1999 and 2000. Acc. 0169, Acc
-        1999-044, Acc 2000-012, Acc 2000-023.</p>
-    </acqinfo>
```
jkeck: We noticed this existed already in the fixture (but another `acqinfo` was added to satisfy the needs for this ticket).  

@anarchivist cool that we delete the original `acqinfo`?

lib/generators/arclight/templates/catalog_controller.rb
```diff
@@ -137,9 +137,16 @@ class CatalogController < ApplicationController
...
-    
+
+    config.show.metadata_partials = [
+      :summary_field,
+      :access_field,
+      :background_field,
+      :scope_and_arrangement_field,
+      :related_field,
+      :admin_info_field
+    ]
```
jkeck: @mejackreed I noticed that we had to add a key to both `Arclight::Engine.config.catalog_controller_field_accessors` and `Blacklight.config.show.metadata_partial` to get things working.  Would it make sense for use to assign `Blacklight.config.show.metadata_partial` to `Arclight::Engine.config.catalog_controller_field_accessors` here, or is it likely that these will diverge in the near future?

mejackreed: Are you proposing moving this config to `engine.rb`? Is so, I would vote no. This is the configuration that the design calls for, but what if I wanted to add or remove something? I guess you could do it, but it would be a bit hidden to do so

jkeck: I'm asking if changing 
```ruby
config.show.metadata_partials = [...]
```
to
```ruby
config.show.metadata_partials = Arclight::Engine.config.catalog_controller_field_accessors
```
would make sense so you don't have to add the same symbol to both arrays in both files as part of the process of adding a new section (this might just be particularly obvious to me since we just had to do that multiple times for this PR).

That's why I asked the question re: the likelihood of these two arrays diverging in the near future (which it's sounding like you believe they will).

mejackreed: Ah.. yes I believe they will. At some point we will need to change:
```diff
- config.show.metadata_partials = [...]
+ config.show.collection.metadata_partials = [...]
+ config.show.component.metadata_partials = [...]
```
Or something like that

jkeck: As it stands, I don't see why you would add a field to the `catalog_controller_field_accessors` that would not result in an additional metadata section, but you have obviously thought through this particular feature more than I.

mejackreed: Ah its based on display_type. So for the component mocks, they show different MD sections.

jkeck: Makes sense, @drh-stanford and I were just talking about how all of this stuff would most likely need to be namespaced to the collection once we start building out the views for components.

PR #182, opened 2017-04-21
Add an Indexed Terms section to the Collection show page.

### LINKED ISSUE #179: Add indexed terms to collection show page
Broken out of #144 

* Indexed terms
  * Specific children of `<controlaccess/>`: `<corpname/>`, `<famname/>`, `<function/>`, `<genreform/>`, `<geogname/>`, `<occupation/>`, `<persname/>`, `<subject/>`, and `<title/>`.

### LINKED ISSUE #184: Index controlled access terms (esp. subjects) at the current level only
In [issue #11 (comment)](https://github.com/sul-dlss/arclight/issues/11#issuecomment-295937041), @jkeck called out the following:

> The xpath in the current PR indexes any control access sub-elements into the collection (even when they are elements in a component).

We should actually be indexing controlled access terms only at the level in which they appear, e.g. a collection should only have controlled access terms associated with those appear at the collection level, and a component should only have those associated with the component itself.

jkeck:
Closes #179
Closes #184 

## Collection object

indexed-terms

## Faceting (#184)
subject-facets

### DIFF
--- a/app/presenters/arclight/show_presenter.rb
+++ b/app/presenters/arclight/show_presenter.rb
@@ -9,5 +9,22 @@ module Arclight
       title += ", #{date.first}" if date.present?
       title
     end
+
+    def with_field_group(group)
+      self.field_group = group
+      self
+    end
+
+    private
+
+    def field_group
+      @field_group || 'show_field'
+    end
+
+    attr_writer :field_group
+
+    def field_config(field)
+      BlacklightFieldConfigurationFactory.for(config: configuration, field: field, field_group: field_group)
+    end
   end
 end
--- a/app/factories/blacklight_field_configuration_factory.rb
+++ b/app/factories/blacklight_field_configuration_factory.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+##
+# A factory to return blacklight field configurations given a field key
+# that represents a field group (e.g. summary_fields, access_fields, etc)
+class BlacklightFieldConfigurationFactory
+  def self.for(config:, field:, field_group:)
+    new(config: config, field: field, field_group: field_group).field_config
+  end
+
+  def initialize(config:, field:, field_group:)
+    @config = config
+    @field = field
+    @field_group = field_group
+  end
+
+  def field_config
+    return null_field unless config.respond_to?(:"#{field_group}s")
+    config.send(:"#{field_group}s").fetch(field) { null_field }
+  end
+
+  private
+
+  attr_reader :config, :field, :field_group
+
+  def null_field
+    Blacklight::Configuration::NullField.new(field)
+  end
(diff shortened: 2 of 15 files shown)


jkeck:
@anarchivist, also I'm assuming that there is no requirement to deal with subject hierarchy for the scope of this ticket?

I see things about splitting subjects on `|z` and `--`, which I don't see in our fixture data.  The kind of hierarchy indicated by those separators is what I believe the discussion around subjects during the design review was about.

anarchivist:
There is no requirement to deal with subject hierarchy in this ticket. Re: the following:

> I see things about splitting subjects on |z and --, which I don't see in our fixture data. The kind of hierarchy indicated by those separators is what I believe the discussion around subjects during the design review was about.

Where did you see this - is this in the indexing code?

jkeck:
Sorry, I should have been more clear, and now that I take a closer look at the code I misspoke slightly.

There is [logic in the Arclight indexing code](https://github.com/sul-dlss/arclight/blob/ed97d562e475e3966a148de74288c57abf8d75b1/lib/arclight/custom_document.rb#L75-L77) that deals with subject parsing (just for facets now, but is re-used for this PR) that replaces `|z` with with `--`.  It **doesn't** split them as I indicated in my comment above.  I was confused because there wasn't any data in our fixture that exhibited that behavior, and I didn't see anything about doing that in a quick look over the original ticket.

Sounds like the desire is to keep the `--` in a single subject for now and not try to indicate any hierarchy, which is 👍

app/presenters/arclight/show_presenter.rb
```diff
@@ -9,5 +9,22 @@ def heading
...
+    private
+
+    def field_group
+      @field_group || 'show_field'
+    end
+
+    attr_writer :field_group
+
+    def field_config(field)
+      BlacklightFieldConfigurationFactory.for(config: configuration, field: field, field_group: field_group)
```
jkeck: Interested in folks' take on this.  This is in response to https://github.com/projectblacklight/blacklight/blob/master/app/presenters/blacklight/show_presenter.rb#L89-L91, which explictly uses `config.show_fields`.  This allows us to pass in a `field_group` that can be used to determine which configuration to use.

Without this, we can't access any configuration options that we would want to use in Blacklight's rendering pipeline such as `separator_options` (used in this PR), `helper_method`, `link_to_facet` etc.

Another option would be to write presenters for each field group, but that felt, eh.

anarchivist:
Yeah, let's not split the hierarchy for now. Thank you!

app/presenters/arclight/show_presenter.rb
mejackreed: 👍

PR #187, opened 2017-04-22
Add a header partial for collection documents.

### LINKED ISSUE #171: Collection detail page: The repository holding the collection and the collection identifier/unitid appears in a single line

jkeck:
Closes #171 

collection-header

### DIFF
--- a/app/views/catalog/_arclight_document_header_collection.html.erb
+++ b/app/views/catalog/_arclight_document_header_collection.html.erb
@@ -1,0 +1,10 @@
+<div class='card document-header'>
+  <div class='card-block row'>
+    <div class='col-9'>
+      <%= document.repository %>
+    </div>
+    <div class='col-3 text-right'>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_arclight_document_header_collection.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class='card document-header'>
+  <div class='card-block row'>
+    <div class='col-9'>
+      <%= document.fetch('repository_ssm', []).first %>
+    </div>
+    <div class='col-3 text-right'>
+      <%= t(:'arclight.views.show.collection_id', id: document.fetch('unitid_ssm', []).first) %>
```
jkeck: Wondering if there's something better we can do that hard-coding these into our views.

mejackreed: Potentially add a method accessor here? https://github.com/sul-dlss/arclight/blob/master/app/models/concerns/arclight/solr_document.rb

PR #195, opened 2017-04-25
Spy of the scroll

### LINKED ISSUE #178: Add scrollspy behavior to sidebar on collection show page
Broken out of #144

- [ ] The navigation aid should also exhibit "scrollspy" behavior, highlighting the currently active section.

mejackreed:
Closes #178 

Adds in a sticky sidebar and enables scrollspying. Bootstrap4 removed affix plugin, so we needed to add our own implementation using `position: sticky`. This causes some issues in some browsers (not current and IE/edge) so also added a polyfill for it.

## Update to add heading
screen shot 2017-04-25 at 8 20 04 am

## Sticky scrollspy on edge

scrollspy

Sources:
https://v4-alpha.getbootstrap.com/migration/#components
https://developer.mozilla.org/en-US/docs/Web/CSS/position#Sticky_positioning
http://caniuse.com/#feat=css-sticky
https://github.com/wilddeer/stickyfill


jkeck:
I think it'll be useful to get this in front of @ggeisler on the demo site.  I think there may be some issues around how the accordion section below will be accessible w/ the sticky sidebar positioning, but I don't think we should hold this up for it. 👍

Issue #188, opened 2017-04-24
Collection detail page overview: link terms/names to new faceted searches 


anarchivist:
On the collection detail page, the following items should be linked to new faceted searches. The location of these elements should stay the same.

* ~Creator (i.e. children of `<origination/>` - `<persname/>`, `<corpname/>`, `<famname/>`): link to a new Names facet query~
* ~Subjects (i.e. `<subject/>`, `<function/>`, `<genreform/>`, `<occupation/>`): link to a new Subject facet query~
* ~Names (i.e. `<corpname/>`, `<persname/>`, `<famname/>` and `<name/>` but only when children of `<controlaccess/>`): link to a new Names facet query~
* ~Places (i.e. `<geogname/>` but only when children of `<controlaccess/>`): link to a new Places facet query~

**EDIT**: #202 implements a link to an "all subjects" facet. #207 addresses the need to link specific terms/names back to the appropriate kind of facet query.

**Mockup**

![arclight-mockups-mvp2](https://cloud.githubusercontent.com/assets/73732/25345031/b3183304-28c8-11e7-9e19-439c94f1b444.jpg)

jkeck:
The grouping of the different subjects to be linked to different facets will require a bit of custom rendering work that may take some additional thought.  Per discussion at standup, we'll take a first pass by simply linking subjects to a single subjects facet, and break out a ticket about linking subjects to separate facets.

cc: @anarchivist

anarchivist:
Does @gordonleacock's PR #202 address the first part? If so, should we just treat this ticket to do that breakout?

anarchivist:
Closed; see #207 for remaining needs.

Issue #149, opened 2017-04-17
Collection detail page - contents/inventory section MVP


anarchivist:
The MVP for the contents section of the collections detail page displays information about the immediate children of the top-level description. 

**Acceptance criteria**

* A navigation bar that allows a user to switch between the summary section and the contents section
* If there are children of the top-level description:
    * There should be an entry for each child of the top-level description, which displays the following information:
        * Title and/or dates as one string, and component identifier (`<unitid/>`) if available, linking to the appropriate component-level show page
        * The first ~175 characters of the scope and content note (if available)
    * The entry in the nav bar should read "Contents"
    * The children should be displayed in order as they appear in the source data -- **CAN BE BROKEN INTO A NEW TICKET**
* If there are no children of the top-level description:
    * The entry in the nav bar should read "No content inventory" and not be clickable

**Out of Scope for this ticket**

* A label indicating the number of children of that child, with the appropriate level if identifiable (#189)

**Mockups**

![collection-detail-contents](https://cloud.githubusercontent.com/assets/73732/25080031/5ffcc360-22f6-11e7-8f7d-96c472d39a79.jpg)

![arclight-mockups-mvp](https://cloud.githubusercontent.com/assets/73732/25080045/7b846070-22f6-11e7-93c8-1d952e0f5ac4.jpg)

mejackreed:
Writing some notes here. I think a successful query might be:

`fq":["component_level_isim:1", "ead_ssi:aoa271"]`

jkeck:
We may want to pull having customized metadata displays for the content inventory into a separate ticket.

PR #202, opened 2017-04-26
issue #188

gordonleacock:
After discussion during standup today (Wednesday 4-26-17) Mark noted that issue #188 was more complicated that just linking to the subjects facet. Linking to the appropriate facet types for subjects, names, and geo is desired.

Current this PR only creates links using the all_subjects_ssim facet. The previously used access_subject_sim has been temporarily disabled in the catalog_controller.rb file.

Jack had suggested to we modified all the _sim and _ssm references to _ssim but I ran into trouble trying that and those changes aren't included here.

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -70,13 +70,15 @@ class CatalogController < ApplicationController
     # :index_range can be an array or range of prefixes that will be used to create the navigation (note: It is case sensitive when searching values)
 
     config.add_facet_field 'collection_sim', label: 'Collection'
-    config.add_facet_field 'creator_sim', label: 'Creator'
+    config.add_facet_field 'creator_ssim', label: 'Creator'
     config.add_facet_field 'date_range_sim', label: 'Date range', range: true
     config.add_facet_field 'level_sim', label: 'Level'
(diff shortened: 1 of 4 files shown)


app/views/catalog/_show_collection.html.erb
```diff
@@ -3,6 +3,14 @@
     <%= render partial: 'show_sidebar' %>
   </div>
   <div class='col-md-9'>
+    <div class="btn-toolbar" role="toolbar" aria-label="Toolbar with button groups">
```
mejackreed: Do you mind pulling these buttons out? I think they will probably be added as part of #149

lib/generators/arclight/templates/catalog_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 class CatalogController < ApplicationController
 
+  include BlacklightRangeLimit::ControllerOverride
```
mejackreed: Do we need to add this in? Does the BlacklightRangeLimit add this automatically?

solr/conf/schema.xml
```diff
@@ -16,10 +16,10 @@
  limitations under the License.
 -->
 
-<!--  
+<!--
```
mejackreed: Is it possible to move these whitespace changes to a separate commit?

PR #203, opened 2017-04-26
Update layout for search results documents

### LINKED ISSUE #190: Style collection-level information (for search results, etc.)
Broken out of #139.

* Each collection-level result displays the following
    * Repository and collection identifier (`unitid`) as a single line
    * Collection title and dates, as a single line
    * Creator
    * Extent (if available)
    * The collection's abstract (if available). If the abstract is longer than 175 characters, truncate it in the display

**Mockup** 

collections-page-search-markedup

### LINKED ISSUE #192: Style component-level information in search results
Suggested by @jkeck after discussing #190, since it might be useful to think through whether these are similar partials.

* Each component-level result displays the following:
    * Repository and collection identifier (`unitid`) as a single line (as in #190)
    * Component title and/or dates, as a single line 
        * Note that title OR dates may appear alone
    * Creator (if available) 
    * Extent (if available)
    * The component's scope and contents note (if available). If the abstract is longer than 175 characters, truncate it in the display

jkeck:
Closes #190 
Closes #192

results

_**Note:** this removes the bootstrap card stying that was previosly applied to the repository/collection ID bar, which is shared on the record view, so that now looks like:_
collection-record


ggeisler:
Looks promising, @jkeck. I think using divs for the metadata is fine (at least, I can't think of a reason now that that isn't sufficient). I also like removing the Bootstrap card styling. The card styling is used a lot with Bootstrap 4 (good overall) but in this case I think avoiding the `card` class will make it easier to style the collection bar appropriately.

The classes on the component levels (`blacklight-subseries`, `blacklight-file`, etc.) seem like they will be fine for styling. Below I colored each distinct class differently (removed some `blacklight-otherlevel` entries for screenshot purposes) just to make sure I could distinguish each level, and it seems fine.

So for styling purposes anyway, this looks good. Hard to say for sure we won't need some adjustments until we really get into the CSS to try to lay things out visually, but I think this definitely gives me enough to get started with.

-_blacklight_search_results

ggeisler:
Oh, as for the Bookmark checkbox, yes, we had ideas for styling it a bit (essentially replacing the checkbox with a bookmark icon), but I have no problem deprioritizing that and just going with the default styling for the time being.

PR #204, opened 2017-04-26
Add an online content indicator badge to collections.

### LINKED ISSUE #148: "Online content" label
If a component or collection has digital objects associated with it that are accessible online, an additional "online content" label should appear in a variety of contexts.

**Acceptance criteria**

* Indexing should recognize the presence of `<dao/>` tags denoting associated digital objects (perhaps a new boolean field?)
* When associated digital objects exist, the "online content" label should appear in the following contexts:
    * Collection detail pages
    * Collection-level entries on contents listings, search results, etc.

**Mockups**

jkeck:
Closes #148 

## Record View
show-oc-indicator

## Search Results
index-oc-indicator

### DIFF
--- a/lib/arclight/custom_document.rb
+++ b/lib/arclight/custom_document.rb
@@ -39,17 +39,25 @@ module Arclight
     def to_solr(solr_doc = {})
       super
       solr_doc['id'] = eadid.first.strip.tr('.', '-')
-      Solrizer.insert_field(solr_doc, 'level', 'collection', :displayable) # machine-readable
-      Solrizer.insert_field(solr_doc, 'level', 'Collection', :facetable) # human-readable
-      Solrizer.insert_field(solr_doc, 'names', names, :facetable)
-      Solrizer.insert_field(solr_doc, 'date_range', formatted_unitdate_for_range, :facetable)
-      Solrizer.insert_field(solr_doc, 'access_subjects', access_subjects, :facetable)
-      Solrizer.insert_field(solr_doc, 'all_subjects', all_subjects, :symbol)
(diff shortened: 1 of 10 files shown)


lib/arclight/custom_document.rb
```diff
@@ -45,6 +45,7 @@ def to_solr(solr_doc = {})
       Solrizer.insert_field(solr_doc, 'date_range', formatted_unitdate_for_range, :facetable)
       Solrizer.insert_field(solr_doc, 'access_subjects', access_subjects, :facetable)
       Solrizer.insert_field(solr_doc, 'all_subjects', all_subjects, :symbol)
+      Solrizer.insert_field(solr_doc, 'has_online_content', online_content?, :displayable)
```
jkeck: Anybody have any good ideas on how to reduce the complexity of this method?

I'm thinking that we could build up an array of hashes representing these fields and just iterate through that array.  Alternatively, we make this the impetus for generating a `.rubocop_todo.yml`.

jkeck: I added e38fd6f92a8e47242a78bdbf42f7359a7801a810 in response to this.  Happy to kill that and generate a `rubocop_todo.yml` file (or put it in the config, but it seems like we would have to exclude the file entirely).

jkeck:
I had thought about going that route, but my concern is that we'll actually want them to live separately in the UI.

It seems like the "actiony" things will be `Bookmark`, `Request This Item` etc, and those will reside in the bar at the top of the record, where this label will end up living separately from those actions from the UI perspective.

Is there a good way to split that sort of thing up in the list of actions to be put in different places in the UI?

mejackreed:
Nah, I didn't think of that, but totally makes sense. 🚢 thanks for that

PR #219, opened 2017-04-28
add facet items to collection summary

gordonleacock:
Still have one rubocop issue: In lib/arclight/custom_document.rb, the "  extend_terminology do |t|" block is too long. Two fixes I can think of:
1. Add file to .rubocop.yml in the "Metrics/BlockLength:   Exclude:" section.
2. Turn all calls with extend_terminology into a method that takes an array of terms, paths, and index_as elements.

Note: We only have one creator child, so it is hard to test that part.
Also, I still need to add rspec tests for this issue!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -31,6 +31,7 @@ Metrics/BlockLength:
     - 'arclight.gemspec'
     - 'spec/**/*'
     - 'tasks/**/*'
+    - 'lib/arclight/custom_document.rb'
 
 Performance/RegexpMatch:
   Enabled: false
--- a/lib/arclight/custom_document.rb
+++ b/lib/arclight/custom_document.rb
@@ -11,9 +11,13 @@ module Arclight
       t.unitid(path: 'archdesc/did/unitid', index_as: %i[displayable])
       t.repository(path: 'archdesc/did/repository/corpname/text() | archdesc/did/repository/name/text()', index_as: %i[displayable facetable])
       t.creator(path: "archdesc/did/origination[@label='creator']/*/text()", index_as: %i[displayable facetable symbol])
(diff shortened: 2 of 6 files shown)


app/views/catalog/_show_collection.html.erb
```diff
@@ -3,6 +3,12 @@
     <%= render partial: 'show_sidebar' %>
   </div>
   <div class='col-md-9'>
+    <div class="btn-toolbar" role="toolbar" aria-label="Toolbar with button groups">
+  <div class="btn-group" role="group" aria-label="First group">
+    <button type="button" class="btn btn-primary">Overview</button>
```
mejackreed: It looks like this made its way into the commit again.. can this be removed?

solr/conf/schema.xml
```diff
@@ -16,10 +16,10 @@
  limitations under the License.
 -->
 
-<!--
```
mejackreed: Can we remove these whitespace changes again?

gordonleacock: Ditto!

jkeck:
@gordonleacock I think excluding the file in the `.rubocop.yml` would be 👌.

A comment about `extend_terminology` requiring a long block (similar to how our `Metrics/LineLength` exclusions).

gordonleacock:
I have a question out to @anarchivist on slack that will enable me to resolve the coverage issue.

anarchivist:
@gordonleacock If I'm understanding the question, I think you're asking if we should undo the "All Subjects" facet in to go back to the previous "Subjets" facet. My answer is yes - we should go bak to the old "Subjects" behavior.

Let me know if you need changes to the fixture data to test things further!

gordonleacock:
@jkeck - I've updated this PR per your suggestion. I hope it can get in for the demo.

PR #228, opened 2017-04-29
Move bookmark control into document title bar …

### LINKED ISSUE #223: Relocate bookmark/request checkboxes in search result items
As discussed at standup, we need to resolve the issue of where to place the bookmark/request checkboxes in relation to the collection/item identifiers and where relevant, the "online content" badge. Currently things like like this:

-_blacklight_search_results

I suggest that we:

* Leave the "online content" badge in the same row at the result item title, as I think it currently is
* Show the collection ID/item hierarchy left-aligned after the parent repository name
* Move the bookmark and request checkboxes to the `.text-right` div in the `.al-document-title-bar` div (I'd also reduce the `.text-right` div to 3-col and increase its sibling div to 9-col)

So the arrangement would look roughly like this (the mockup below includes reducing the font-size in the  `.al-document-title-bar` div and other styling updates, which I can do in a separate PR):

collection__the_italian_or_the_confessional_of_the_black_penitents___typescript_-_blacklight_search_results 2

jkeck:
…(and don't render on collection show).

Closes #223 

## Results
results

## Collection Show
collection-show

## Compontent Show
compontent-show

### DIFF
--- a/app/models/concerns/arclight/solr_document.rb
+++ b/app/models/concerns/arclight/solr_document.rb
@@ -26,6 +26,10 @@ module Arclight
       first('repository_ssm')
     end
 
+    def repository_and_unitid
+      [repository, unitid].compact.join(': ')
+    end
+
     def collection_name
       first('collection_ssm')
     end
--- a/app/views/catalog/_arclight_document_index_header.html.erb
+++ b/app/views/catalog/_arclight_document_index_header.html.erb
@@ -1,0 +1,13 @@
+<% document_actions = capture do %>
+  <% # bookmark functions for items/docs -%>
+  <%= render_index_doc_actions document, wrapping_class: "col-3 text-right" %>
+<% end %>
+
+<div class='al-document-title-bar'>
+  <div class='row'>
+    <div class='col-9'>
+      <%= document.repository_and_unitid %>
+    </div>
+    <%= document_actions %>
+  </div>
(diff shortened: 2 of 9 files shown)


ggeisler:
We might not have been fully on the same page about the Collection Show page. When in my last comment on #223 I said:

> ideally the al-document-title-bar of the Collection show page would not inherit this change, but would remain the way it currently is.

I meant just the way it is now, with the `Collection ID: MS C 271` right-aligned.

I think you interpreted that as just not including the bookmark checkbox?

If that is a big problem I'd say we just go with what's in the PR and decide later if it we really need to change it.

jkeck:
@ggeisler so just to be clear, on the collection show page we would want:

1. `Collection ID: Blah` right aligned.
1. **Not** to render the bookmark control

?

ggeisler:
Yes, the same as in the MVP mockups for the Show Collection page:

arclight-mockups-mvp_pdf__page_8_of_23_

jkeck:
## Updated Collection Show Page

updated-collection-show-page

ggeisler:
Great, thanks!

ggeisler:
I don't think I'm qualified to review these changes, but if @mejackreed is able to by early Monday morning, I should have time to submit a PR with styling updates so things look better for this sprint's demo.

PR #252, opened 2017-05-02
Context needs a document_index_view_type

drh-stanford:
https://github.com/projectblacklight/blacklight/blob/master/spec/helpers/blacklight_helper_spec.rb#L37


drh-stanford:
The build passed but the last `phantomjs --version` is still 1.9.8

screen shot 2017-05-03 at 11 04 31 am

jkeck:
I believe it is running under the correct version of PhantomJS, it's just that the version isn't being recognized by the shell at that time.  I'll try adding a `hash -r` before we print out that version so we know that we have the correct version in the output.

Will just push a commit to this branch.

drh-stanford:
ok this looks great now!

PR #253, opened 2017-05-03
issue232 add component metadata DO NOT MERGE

gordonleacock:


app/views/catalog/_component_overview.html.erb
```diff
@@ -0,0 +1,10 @@
+<% document ||= @document %>
```
jkeck: I don't believe this is necessary since we're using `@document` everywhere below.

gordonleacock: Done.

PR #262, opened 2017-05-03
Change indexing to use config/repositories.yml if a slug is provided by REPOSITORY_ID

### LINKED ISSUE #226: During indexing, specify a configured repository with which to associate indexed collections/components
**Acceptance criteria**

* When indexing, I should be able to _optionally_ specify an identifier/slug for a configured repository (see #222) 
* Indexed documents should have information assigned from the configured repository information rather than data from the source to be indexed
* If I try to index with an identifier/slug that doesn't match a preconfigured one, throw an error

drh-stanford:
This PR fixes #226. The usage is like so:

```
REPOSITORY_ID=slug_abc bundle exec rake arclight:index FILE=my_file.xml
```

or for a directory with **files all for the same repository** (for example, when someone has organized all their files using directories named after the repository -- which is a use case Mark mentioned to me.)

```
REPOSITORY_ID=slug_abc bundle exec rake arclight:index_dir DIR=my_dir
```

Also you can specify `REPOSITORY_FILE=my_dir/repositories.yml` if you're not using the default `config/repositories.yml` location. This is useful for the specs.

One concern I have is that for the document (collection) indexing I had to change `Solrizer.insert_field` to `Solrizer.set_field` to follow the pattern for `arclight_field_definitions` in that class. All the specs pass but I'd like feedback on that change.

I also added some documentation and timings to the `index.rake` and changed it to have `arclight:index_dir` use the `arclight:index` task to do its work.


lib/arclight/custom_document.rb
```diff
@@ -44,13 +44,15 @@ def to_solr(solr_doc = {})
       super
       solr_doc['id'] = eadid.first.strip.tr('.', '-')
       arclight_field_definitions.each do |field|
-        Solrizer.insert_field(solr_doc, field[:name], field[:value], field[:index_as])
+        # we want to use `set_field` rather than `insert_field` since we may be overriding fields
+        Solrizer.set_field(solr_doc, field[:name], field[:value], field[:index_as])
```
jkeck: I think this is 👍.  I believe in every case we've been assuming we are overriding the field entirely.

lib/arclight/custom_component.rb
```diff
@@ -28,11 +28,20 @@ def to_solr(solr_doc = {})
...
+      resolve_repository(solr_doc)
       solr_doc
     end
 
     private
 
+    attr_accessor :repository
+
+    def resolve_repository(solr_doc)
+      @repository ||= solr_doc['repository_ssm']
```
jkeck: I have to admit, I had to read this over a few times to grok what was going on here.  The setting of the accessor that's used in the mixin adds a bit to the cognitive load of this code.

What would you think about passing the value as a parameter to the `repository_as_configured` method?

jkeck:
Are there any considerations we'll need to make for indexing in the demo app?

drh-stanford:
Yes, for the demo, I imagine that we're going to want to have a few/several pre-configured repositories.yml repos. Then we'll need to change the `demo:seed` rake task over there to send in the `REPOSISTORY_ID` values for each file. @anarchivist mentioned that NYU(?) uses a directory structure to organize their EADs by repository and the directory name is the REPOSITORY_ID. We could do something like that for the demo's `data/ead` folder.  We probably need a new ticket for that work...

lib/arclight/custom_component.rb
drh-stanford: Yeah, I struggled with this piece of the code. I put `repository_as_configured` as a shared method with no parameters which yielded this requirement for having a repository accessor. The document already has one (defined by the terminology) but the component does not. So that's why it's like that on lines 37-40.

I can parameterize that shared function if it helps.

drh-stanford:
Created #268 and #269 tickets for the demo site work

Issue #222, opened 2017-04-28
Structured repository information and repositories page


anarchivist:
**Subtasks**

- [x] Support for file-backed `Repository` class - see #251
- [x] Load configuration file via generator - see #251
- [x] Add header for Repositories - see #260
- [x] Page to display all repositories - see #260
- [x] Add breadcrumbs - see #260
- [x] Update styling - see #290 
- [ ] Linkify repository name - see #295 
- [ ] Linkify contact email address (mailto link) - see #295 

**Acceptance criteria**

* Ability to define structured information about a repository 
    * Note: I'm neutral on the specific implementation; admin UI mockups suggest this should be admin user editable within that interface. This could be file-backed for now.
    * Identifier or slug (required)
    * Repository name (required)
    * Location
         * Building
         * Address 1
         * Address 2
         * City
         * State
         * ZIP
         * Country
    * Contact phone number
    * Contact information
    * Description
    * Thumbnail (just a URL to the thumbnail is fine for now)
* Display a repositories landing page
    * Display shared application header
         * Indication that the "Repositories" page is active should be clear
    * Display all repositories defined
    * For each repository, display all structured information as defined above [this will be a partial used in other contexts]
    * A route for the page (e.g. `/repositories`)
    * Breadcrumb link leads back to the application home page

**Out of scope**

* Count of collections
* "Learn more" link

**Mockup**

![repositories landing](https://cloud.githubusercontent.com/assets/73732/25540772/1662a0bc-2c01-11e7-869a-5a133bc604c1.jpg)

drh-stanford:
Do we want to store this information in a hand-edited `repository.yml` file only? Or should we be using an ActiveRecord model `Arclight::Repository` and use `db/seeds.rb` to load the data into the database?

I'm assuming that we'll need to do the latter eventually if we have an Admin UI that can customize these data. If that's a correct assumption, should we take the time to do a database-backed implementation now?

jkeck:
I wonder if we could get away w/ a PORO that reads from a `repository.yml` that has an ActiveRecord like interface (maybe implements `where` and `all`) to allow use to easily swap out.

This way we don't need to worry about seeds, schema, etc until we're actually ready to make the records creatable via a UI.

drh-stanford:
Ok, that sounds good. So I'll create a Repository data class that has a `from_yaml` and a couple finder methods and that should lend its way to migrating to ActiveRecord if we need to.

drh-stanford:
[...]

PR #274, opened 2017-05-05
Add Repository card to search results when faceted on repository

### LINKED ISSUE #225: Display repository-specific information on top of search results faceted by repository
Prerequisites: #222, #226  

**Acceptance criteria** 

* Assumptions
    * At least one repository configured according to data structure laid out in #222 
    * At least one finding aid associated with a configured repository (see #226)
* On a search result page faceted by repository:
    * Display the existing facet info/clear info
    * If the repository is configured using the #222 data structure,
         * display the appropriate "card" display containing repository info
    * Otherwise, don't display anything additional

**Mockup**

arclight-mockups-mvp

drh-stanford:
This PR fixes #225. It adds a repository card when needed in the search results.

The PR adds a new finder for the Repository class:

```
repository = Arclight::Repository.find_by(name: 'My Repository')
repository = Arclight::Repository.find_by(slug: 'my-slug')
```

The PR does a few improvements on the repository test data too:
- Updates the `spec/fixtures/config/repositories.yml` with more realistic data
- Reorganizes the EAD fixture data to use repository slugs as directory names (ala the demo site)
- Indexes the seed data with the repositories.yml data (note that lib/generators has the real repo data and we might want to just use the template there and install the real repo data via the test_app_templates later)

screen shot 2017-05-05 at 1 03 37 pm


lib/arclight/repository.rb
```diff
@@ -59,11 +59,18 @@ def self.all
...
+    # Mimics ActiveRecord dynamic `find_by` behavior for the slug or name
     #
-    # @param [String] `slug`
-    def self.find_by(slug)
-      all.select { |repo| repo.slug == slug }.first
+    # @param [String] `slug` or `name`
+    # @return [Repository]
+    def self.find_by(slug: nil, name: nil)
+      if slug
+        all.select { |repo| repo.slug == slug }.first
```
jkeck: Can we use `find` here instead? (and then we can drop `first` at the end).  I'm surprised rubocop doesn't complain about this.

app/helpers/arclight_helper.rb
```diff
@@ -43,6 +43,14 @@ def fields_have_content?(document, field_accessor)
     end
   end
 
+  # If we have a facet on the repository, then return the Repository object for it
+  #
+  # @return [Repository]
+  def repository_faceted_on
+    repos = facets_from_request.find { |f| f.name == 'repository_sim' }.try(:items)
+    Arclight::Repository.find_by(name: repos.first.value) if repos.length == 1 && repos.first.value
```
jkeck: If `try` fails here `repos` will be `nil`.  But we call `length` and `first` on repos on this line.  Could this be potentially problematic if we don't find a repository?

jkeck:
Hmm, now that I look at the mockups, I'm not sure the "cards" from the repositories page is the exact same markup/layout that we'll want for the search results.  Perhaps we can get some clarification from @ggeisler.

ggeisler:
@jkeck @drh-stanford Yes, good eye Jessie. In the search results context it doesn't make as much sense to show the collection count and the learn more button so we left those off (p. 16 of the MVP set of mockups). It's a subtle thing but probably makes sense for the card to be slightly different here.

lib/arclight/repository.rb
drh-stanford: My understanding is that `find` is to grab the record by its "ID" value. `find_by` and `where` were the two options I found for a way to search by name or by slug. Or are you suggesting that we implement a `find` that looks for matches in both the slug and name data?

app/helpers/arclight_helper.rb
drh-stanford: I can add that to the guard. I think it'd only happen if the repository facet was empty though, right? Maybe there's a smarter way to detect if we're faceted by repository?

drh-stanford:
Ok, then it sounds like we can't reuse the partial for the repository card as-is. I'll update it...

lib/arclight/repository.rb
jkeck: I mean using `all.find { |repo| repo.slug == slug }` instead of `all.select { |repo| repo.slug == slug }.first`

`select` iterates through the entire array that you call it on, and `find` will find the first and exit, which would make sense since we're simply calling `first` anyway.

jkeck: `detect` could also be used (as they are the same method).

https://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-find

app/helpers/arclight_helper.rb
```diff
@@ -43,6 +43,17 @@ def fields_have_content?(document, field_accessor)
     end
   end
 
+  # If we have a facet on the repository, then return the Repository object for it
+  #
+  # @return [Repository]
+  def repository_faceted_on
+    if try(:search_state) # rubocop: disable Style/GuardClause
```
jkeck: If we want to follow the style guide w/o doing an inline disable here, we could add `return unless try(:search_state)` at the top of the method (which should be functionally equivalent).

lib/arclight/repository.rb
drh-stanford: ok gotcha

drh-stanford:
ok, i believe i've addressed all the comments. thanks for the feedback. a note on the styling -- #222 still has a subtask to improve the styling, so I didn't do that in this PR.

PR #275, opened 2017-05-05
Ensure the eadid is cleaned up when dealing with the parent data.

### LINKED ISSUE #264: Breadcrumbs: broken component links with collections with an eadid containing periods
For example

* `ARS.0046` is the value of the `<eadid/>` in this screen recording.
* The "Instantaneous Recordings" component has the (correctly rewritten) link `http://localhost:3000/catalog/ARS-0046aspace_ref16_dvr`
* The breadcrumb item in "Box 2: Decca/Gramophone/Odeon/Instantaneous" for "Instantaneous Recording" has the incorrect link `http://localhost:3000/catalog/ARS.0046aspace_ref16_dvr`

rfolprus9m

jkeck:
Fixes #264

### DIFF
--- a/app/models/arclight/parents.rb
+++ b/app/models/arclight/parents.rb
@@ -13,7 +13,7 @@ module Arclight
     end
 
     def eadid
-      (@eadid || '').tr('.', '-')
+      Arclight::NormalizedId.new(@eadid).to_s
     end
 
     ##
--- a/app/models/arclight/parents.rb
+++ b/app/models/arclight/parents.rb
@@ -5,11 +5,15 @@ module Arclight
   # Object for parsing and formalizing Solr_Ead "Parents"
   # https://github.com/awead/solr_ead/blob/8cf7ffaa66e0e4c9c0b12f5646d6c2e20984cd99/lib/solr_ead/behaviors.rb#L54-L57
   class Parents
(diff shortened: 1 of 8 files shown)


app/models/arclight/parents.rb
```diff
@@ -9,7 +9,7 @@ class Parents
     def initialize(ids:, labels:, eadid:)
       @ids = ids
       @labels = labels
-      @eadid = eadid
+      @eadid = eadid.tr('.', '-')
```
drh-stanford: i've seen this logic in a few places. we could add a `normalize_ead` function or something for it. 

see https://github.com/sul-dlss/arclight/blob/ed73b69696dd5f33ae8835b6fc9f802a9d1ea10f/lib/arclight/solr_ead_indexer_ext.rb#L29

https://github.com/sul-dlss/arclight/blob/3d911d0220ef7a00b082d45ec0cd242a650446de/lib/arclight/custom_document.rb#L45

https://github.com/sul-dlss/arclight/blob/3d911d0220ef7a00b082d45ec0cd242a650446de/lib/arclight/custom_component.rb#L27

jkeck: 👍 Yeah, I was thinking the same thing when I was writing this.  Glad you thought that too, I'll add something.

PR #280, opened 2017-05-05
cleanup title/date headings

### LINKED ISSUE #273: Properly render components that contain unitdate without unittitle
Per archival description practices, component levels should have a title (`<unittitle/>` in EAD), a date (`<unitdate/>`), or both. Currently, components with only `<unitdate/>`s are not displaying properly (see screenshots).

**Screenshots**

screen shot 2017-05-04 at 10 01 54 am

screen shot 2017-05-04 at 2 45 14 pm

**Acceptance criteria**

* A component with a `<unittitle/>` and a `<unitdate/>` should be rendered as it is now
* A component with a `<unitdate/>` only should display the data within the `<unitdate/>`'s text node 
* This change should occur whenever a component title gets rendered (including breadcrumbs, if applicable)

**Sample data** 

* [umich-bhl-0312.xml](https://drive.google.com/open?id=0B3DqcSBOaZWWMEJqZTVTQ3ZXYlE)

drh-stanford:
This PR fixes #273.

### DIFF
--- a/app/presenters/arclight/show_presenter.rb
+++ b/app/presenters/arclight/show_presenter.rb
@@ -4,9 +4,7 @@ module Arclight
   # Custom presentation methods for show partial
   class ShowPresenter < Blacklight::ShowPresenter
     def heading
-      title = super
-      delimiter = heading_delimiter(title)
-      view_context.safe_join([title, document.unitdate].compact, delimiter)
+      view_context.safe_join([normalize_title(super), document.unitdate].compact.map(&:html_safe), ', ')
     end
 
     def with_field_group(group)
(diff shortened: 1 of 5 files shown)


jkeck:
@drh-stanford did you ever get an answer to [your question](https://github.com/sul-dlss/arclight/issues/273#issuecomment-299588227) about if there wasn't a title or a date?

drh-stanford:
Nope, this PR assumes that there's either a date or a title, but won't work if there's neither.

jkeck:
@anarchivist can you shed any light on this possibility and what we should do in that case?  I worry about creating the ability where we will have search results documents that don't have any link to their show view.

anarchivist:
As I said on #273:

> If there's no title or date, than using the `id` is fine, but a component "should" have at least a date OR a title. The data is bad if there's not at least one or the other, but displaying the ID should be a good way to indicate that.

app/presenters/arclight/show_presenter.rb
```diff
@@ -5,6 +5,7 @@ module Arclight
   class ShowPresenter < Blacklight::ShowPresenter
     def heading
       title = super
+      title = nil if title.blank? || title == @document.id.to_s
```
jkeck: Can we return the id if there is no date or title? per https://github.com/sul-dlss/arclight/pull/280#issuecomment-299603090

jkeck: Also, can we use `document` instead of `@document` to be more consistent with the other usage in this method?

drh-stanford:
I've updated the branch to do:

- title and date present, join with a comma
- title or date present, use the one that's present
- title and date not present, use the document id

drh-stanford:
Thanks for the feedback. This PR is ready to review now.

PR #302, opened 2017-05-09
Add online section to component sidebar

### LINKED ISSUE #237: Display link/description for digital objects
Prerequisite: (PR: #253)

**Acceptance criteria**

Display links, and when available, descriptions of digital objects at either the collection and component level show views. For this ticket, the understanding is that a `<dao/>` tag must have an `href` attribute containing a valid URI, and may occur within the top-level `<archdesc/>` tag or individual `<c/>` components.

* In the left sidebar add a new section titled "Online" when there is a `<dao/>` at the appropriate level.
* In this section, for each `<dao/>` at a given level:
    * Display either the `<dao/>`'s `title` attribute OR its `<daodesc/>` 
    * Add a ink to the URI identified in the `href` link with the text "Open"

jkeck:
Closes #237 

## Collection (aoa271)
online-collection-sidebar

## Component (aoa271aspace_843e8f9f22bac872d0802d6fffbb04)
online-sidebar

### DIFF
--- a/app/assets/stylesheets/arclight/modules/sidebar.scss
+++ b/app/assets/stylesheets/arclight/modules/sidebar.scss
@@ -1,3 +1,12 @@
 .al-in-person-repository-location {
   clear: left;
 }
+
+.al-digital-object {
+  margin-bottom: $spacer;
+}
+
+// Hiding this from display since we don't need it.
+dt.blacklight-digital_objects_ssm {
+  display: none;
+}
--- a/app/controllers/concerns/arclight/field_config_helpers.rb
+++ b/app/controllers/concerns/arclight/field_config_helpers.rb
@@ -8,11 +8,21 @@ module Arclight
 
     included do
       if respond_to?(:helper_method)
+        helper_method :context_sidebar_digital_object
         helper_method :repository_config_present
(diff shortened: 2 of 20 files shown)


app/assets/stylesheets/arclight/modules/sidebar.scss
```diff
@@ -1,3 +1,12 @@
...
   clear: left;
 }
+
+.al-digital-object {
+  margin-bottom: $spacer;
+}
+
+// Hiding this from display since we don't need it.
+dt.blacklight-digital_objects_ssm {
+  display: none;
```
jkeck: Wasn't quite sure what to do here.  This goes w/ Blacklight's default rendering and then hides the label that we don't want w/ `display: none`.  We could do a screen-reader only approach, but it wasn't obvious that this label was actually necessary.

We could also just do something completely custom, but then coordinating which accordion section should be open first might get a bit ugly.

ggeisler: I'd say go with this for now and at some point soon I'll look at the page again from an accessibility perspective and adjust if necessary.

PR #303, opened 2017-05-09
Update home masthead and add masthead to other pages

### LINKED ISSUE #297: Update masthead
Per 5/8/17 discussion in Slack with @jkeck and @ggeisler:

**Acceptance criteria:**

* On the home page, update the masthead to contain the "Repositories" and "Collections" links
* On all other pages, insert the same configurable heading ("Archival Collections at Institution") between the top menu bar and the row containing the "Repositories"/"Collections" links and search bar

**Mockups** 

arclight-mockups-mvp

arclight-mockups-mvp

ggeisler:
Fixes #297.

One TODO noted in the `mastheads.scss` file is to figure out why the toggler icon from Bootstrap is not (seemingly) being loaded in the masthead, although it is in the topbar. This is pretty minor and I'll figure it out when improving other responsive issues.

Examples:

### Home, updated
blacklight 2
---
### Repositories
blacklight

---
### Search result
within_collection__material___collection__alpha_omega_alpha_archives_-_blacklight_search_results
---
### Mobile
-_blacklight_search_results

### DIFF
--- a/app/views/catalog/_home.html.erb
+++ b/app/views/catalog/_home.html.erb
@@ -1,11 +1,1 @@
-<div class='jumbotron'>
-  <div class='row'>
-    <div class='col-md-8 offset-md-2'>
-      <h1 class='text-center'><%= t('arclight.views.home.heading') %></h1>
-    </div>
-  </div>
-  <div class='row'>
-    <div class='col-md-6 offset-md-3'>
-        <%= render_search_bar %>
-    </div>
-</div>
+
--- a/app/assets/stylesheets/arclight/application.scss
+++ b/app/assets/stylesheets/arclight/application.scss
@@ -1,10 +1,11 @@
 @import 'bootstrap/variables';
 @import 'bootstrap_overrides';
 @import 'variables';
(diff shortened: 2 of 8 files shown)


app/views/catalog/_home.html.erb
```diff
@@ -1,11 +1 @@
...
-    <div class='col-md-8 offset-md-2'>
-      <h1 class='text-center'><%= t('arclight.views.home.heading') %></h1>
-    </div>
-  </div>
-  <div class='row'>
-    <div class='col-md-6 offset-md-3'>
-        <%= render_search_bar %>
-    </div>
-</div>
+
```
jkeck: Not saying that we should do anything w/ this PR, but just wondering if we have/should create a ticket about adding content to the home page, just so it doesn't get lost.

Even if we don't get to adding the real content there, we might want to add some instructional information similar to what Blacklight does out of the box.  Not sure what folks thoughts are on that.

I can bring it up during standup, and don't think it should hold up this PR.

Issue #267, opened 2017-05-04
Indexing performance is incredibly poor even on small-ish finding aids


anarchivist:
I was testing indexing some finding aids from the Bentley ([Google Drive folder](https://drive.google.com/open?id=0B3DqcSBOaZWWbmFDYlVyRnB2Q2c)), and even some relatively small finding aids index slowly. 

The finding aid in question is `umich-bhl-0728.xml`.

We ran into this before with finding aids from SPEC, too, and I'll confirm with a specific example. 

See #250 for one known issue.

anarchivist:
I'll add that Bentley's `umich-bhl-0234.xml` is roughly 3x larger, and indexes quickly.

mejackreed:
It might be good to add some benchmarking of the problem here so that we can determine when performance changes.

drh-stanford:
https://github.com/sul-dlss/arclight/pull/262 adds timings to each file indexed so collecting the benchmarking should be easier when that it merged. In the meantime if you prefix your rake task with `time bundle exec rake arclight:index...` you can get timings

anarchivist:
Goals:

* Set up tooling for benchmarking at a more detailed level
* List of EAD files with timings and characteristics
* Identify ideal timings based size/characteristics
* Index larger EAD files for testing

drh-stanford:
Here's what a profile looks like, and most are like this for the longer indexing times (this particular one is from `umich-bhl-2013066.xml`). At a high level, the indexing is taking most of its time inside OM when it's trying to parse the document and each component. There's an n x m problem here because it extracts the entire terminology from each component. It also applies the terminology using `//` as the xpath prefix so it's scanning all the nodes for all the components and all the terminologies. That's why we're seeing ~100k `xpath` calls. OM's xpath "mangler" is complicated and it's the biggest resource consumer. 

```
  %   cumulative   self              self     total
 time   seconds   seconds    calls  ms/call  ms/call  name
  6.11    15.26     15.26   141954     0.11     0.43  OM::XML::TermXpathGenerator.generate_xpath_with_indexes
  4.27    25.92     10.66   158810     0.07     0.13  Nokogiri::XML::Searchable#extract_params
  3.98    35.86      9.94   304440     0.03     0.08  OM.destringify
  3.85    45.48      9.62   183572     0.05     0.10  OM::XML::Terminology#retrieve_term
  3.49    54.19      8.71   592816     0.01     1.09  Array#each
[...]

PR #304, opened 2017-05-09
warn if EAD has unitdates with multiple date ranges

drh-stanford:
Related to #267 and #306. The PR fixes a problem where the indexer would stall and not finish with EADs with more complicated dates than YYYY or YYYY/YYYY. This piece of code is for the blacklight range limit data -- it has to generate year ranges for the facet (e.g., 1999/2001 --> 1999, 2000, 2001). The specs support some of the formats in #306.

This PR just causes the indexer to output a warning if we have multiple date ranges, and it takes the first date range. I decided to just have the warning and pick the first one for now so we can index the EADs that have the ranges right now as-is, and #305 will correct this problem.

Example output:

```
Loading spec/fixtures/ead/umich-bhl-0047.xml into index...
WARNING: Unsupported multi-date data: ["1910/2000", "1997/2000"]
Indexed spec/fixtures/ead/umich-bhl-0047.xml (in 0.481 secs).
```

### DIFF
--- a/lib/arclight/shared_indexing_behavior.rb
+++ b/lib/arclight/shared_indexing_behavior.rb
@@ -5,14 +5,33 @@ module Arclight
   # A mixin intended to share indexing behavior between
   # the CustomDocument and CustomComponent classes
   module SharedIndexingBehavior
+    # TODO: probably should have a proper DateRange class at some point
+    def to_year_from_iso8601(date)
+      return if date.blank?
+      date.split('-').first[0..3].to_i
+    end
+
+    # @param [String] `dates` YYYY or YYYY/YYYY formats, including YYYY-MM, YYYY-MM-DD, and YYYYMMDD
+    # @return [Array<String>] all of the years between the given years
+    # TODO: probably should have a proper DateRange class at some point
+    def to_date_range(dates)
+      return if dates.blank?
+      start_year, end_year = dates.split('/').map { |date| to_year_from_iso8601(date) }
+
+      return [start_year.to_s] if end_year.nil?
+      raise "Unsupported date formats: #{dates}" if (end_year - start_year).abs > 2100
+      (start_year..end_year).to_a.map(&:to_s)
+    end
+
     # @see http://eadiva.com/2/unitdate/
-    # Currently only handling normal attributes and YYYY or YYYY/YYYY formats
+    # @return [Array<String>] all of the years between the given years
     def formatted_unitdate_for_range
       return if normal_unit_dates.blank?
-      normal_unit_date = Array.wrap(normal_unit_dates).first
-      start_date, end_date = normal_unit_date.split('/')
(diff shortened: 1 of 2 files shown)


anarchivist:
See also #305 and #306

lib/arclight/shared_indexing_behavior.rb
```diff
@@ -5,14 +5,33 @@ module Arclight
...
+    # @return [Array<String>] all of the years between the given years
     def formatted_unitdate_for_range
       return if normal_unit_dates.blank?
-      normal_unit_date = Array.wrap(normal_unit_dates).first
-      start_date, end_date = normal_unit_date.split('/')
-      return [start_date] if end_date.blank?
-      (start_date..end_date).to_a
+
+      all_dates = Array.wrap(normal_unit_dates)
+      puts "WARNING: Unsupported multi-date data: #{normal_unit_dates}" if all_dates.length > 1 # rubocop: disable Rails/Output, Metrics/LineLength
```
jkeck: Hmm, I know this might get complicated, but could we use a configured logger here?  I'm guessing the main problem is we don't have access to rails here, but using `puts` feels wrong.  I wonder if we can access the engine configuration and get a class to log to (and fallback to stdout when the configured class isn't available).

Feel free to punt on that, but it just jumped out at me as we rarely see `puts` in checked-in code (and maybe this is how the whole `MediaShelf::Logger` came about in old-hydra stuff).

lib/arclight/shared_indexing_behavior.rb
```diff
@@ -5,14 +5,33 @@ module Arclight
   # A mixin intended to share indexing behavior between
   # the CustomDocument and CustomComponent classes
   module SharedIndexingBehavior
+    # TODO: probably should have a proper DateRange class at some point
+    def to_year_from_iso8601(date)
+      return nil if date.blank?
```
jkeck: `return nil` is a bit redundant here.  As in the `to_date_range` method we can just `return`.  I have a slight preference towards not explicitly returning `nil` and just returning, but I think we should be consistent between the two methods.

lib/arclight/shared_indexing_behavior.rb
drh-stanford: Yeah, I tried using `Rails.logger` but that's undefined in the indexing code. The `arclight:index` task also writes to stdout. Not sure what to do here...

lib/arclight/shared_indexing_behavior.rb
drh-stanford: ok i fixed that.

lib/arclight/shared_indexing_behavior.rb
jkeck: Totally, just my initial reaction.  🏈

Issue #299, opened 2017-05-08
Context display: partial collection hierarchy (ancestors and current component only)


anarchivist:
Broken out of #240. 

**Acceptance criteria**

* [x] Display the current component and its ancestors following a similar pattern as #227

**Out of scope**

* Sibling display #300
* Current component higlighting 

**Mockup**

![arclight-mockups-mvp](https://cloud.githubusercontent.com/assets/73732/25829672/87cf2064-340c-11e7-8911-d9423c1deae6.jpg)

PR #309, opened 2017-05-10
Issue298

gordonleacock:
With some great help from @mejackreed and @jkeck I think this PR is ready.

It uses js Ajax to fill in the component page collection context section (but not yet the rest of the hierarchy that issue #299).


jkeck:
Looks great @gordonleacock.  Looks like there is a conflict w/ some things that have been merged in.  Mind having a look at that?

PR #321, opened 2017-05-12
Add the collection creator to a display field for components

### LINKED ISSUE #310: Index collection creator(s) for all components
Broken out of #294 and #285.

**Acceptance criteria**

* A component should have the creator information for its collection indexed into a new field.

jkeck:
Closes #310 

Note, this doesn't add anything to the Collection Context section because Arclight renders that data from the collection record itself (and we'll get that for free when we add the creator there).

## Request Form
collection-creator-form

### DIFF
--- a/app/models/arclight/requests/google_form.rb
+++ b/app/models/arclight/requests/google_form.rb
@@ -7,7 +7,7 @@ module Arclight
     # query parameters from the form mapping configuratino
     class GoogleForm
       attr_reader :document, :presenter, :document_url
-      delegate :collection_name, :eadid, :containers, to: :document
+      delegate :collection_name, :collection_creator, :eadid, :containers, to: :document
 
       ##
       # @param [Blacklight::SolrDocument] document
--- a/app/models/concerns/arclight/solr_document.rb
+++ b/app/models/concerns/arclight/solr_document.rb
@@ -57,6 +57,10 @@ module Arclight
       first('creator_ssm')
     end
 
+    def collection_creator
+      first('collection_creator_ssm')
+    end
+
     def online_content?
       first('has_online_content_ssm') == 'true'
     end
--- a/lib/arclight/solr_ead_indexer_ext.rb
+++ b/lib/arclight/solr_ead_indexer_ext.rb
@@ -12,6 +12,8 @@ module Arclight
 
       add_count_of_child_compontents(node, solr_doc)
 
+      add_collection_creator_to_component(node, solr_doc)
+
       solr_doc
     end
 
@@ -50,6 +52,14 @@ module Arclight
 
     def add_count_of_child_compontents(node, solr_doc)
       solr_doc[Solrizer.solr_name('child_component_count', type: :integer)] = node.xpath('count(c)').to_i
+    end
+
+    # This mimics similar behavior in Arclight::CustomDocument
+    def add_collection_creator_to_component(node, solr_doc)
+      field_name = Solrizer.solr_name('collection_creator', :displayable)
+      repository = solr_doc[Solrizer.solr_name('repository', :displayable)]
+      creators = node.xpath('//archdesc/did/origination[@label="creator"]/*/text()').map(&:text)
+      solr_doc[field_name] = creators - [repository]
     end
   end
 end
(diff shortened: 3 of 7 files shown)


anarchivist:
Can we get  a field label that reads `Collection Creator` or something similar here?

jkeck:
@anarchivist do you mean in the search results?

As of now, we haven't labeled any of our metadata in the search results.  How would this relate to how we render the creator at the collection level (which is the pattern that this follows).

Would you expect that this would be the only labeled piece of metadata in the search results?  If so, should this be a `<dl>` (I guess a question for @ggeisler).

ggeisler:
The idea of a field label also occurred to me when looking at the PR earlier, but I thought we could wait until after the PR was merged and I could look at things in a fuller context.

FWIW, here are a couple of formatting possibilities, with and without the field label:

### With a field label
level__collection_-_blacklight_search_results

### Without
level__collection_-_blacklight_search_results 2

So I think it might work without a label, just styling it a bit so it stands out different. Not saying every user will know it is the creator; that might depend on the collection and how clear the creator name is. 

So I'm on the fence. Maybe just go with it as is, but we might revisit and ask for it to be turned into a `dl`?

jkeck:
@ggeisler I'm 👍 with that approach.  It certainly won't be difficult to turn it into a `<dl>`.

Should this styling apply to only the creator in the component, or should we also style it similarly when we print the creator on the collection itself?

ggeisler:
@jkeck By "on the collection itself" do you mean the Collection Detail page?

anarchivist:
@ggeisler Just to be clear, this is also trying to make it clear that this is the creator for the collection, and not necessarily the component's creator.

ggeisler:
@anarchivist Hmm, so you're saying the label (if we added one) might be more appropriately "Collection creator"?

Is collection creator appropriate/necessary for every component in the search results, I wonder? Doesn't seem like crucial information in a result snippet to me.

anarchivist:
@ggeisler Right - the intent with the label so someone doesn't incorrectly infer that the individually is the creator of the component. 

I'm not sure if it's appropriate or necessary for every component. Can you talk more about specific concerns you might have here?

ggeisler:
@anarchivist Well, just that this discussion led me to look at what we did in the mockups, and I noticed we didn't include the collection creator as part of the search result snippets at all. Which led to think about how important collection creator is in the search result snippet (obviously it makes sense for the request form).

In general I think we want to keep the result snippets as minimal as possible, while including information that will help the user determine if a given snippet is potentially relevant to their need. I'm questioning whether "collection creator" is useful enough in determining potential relevance to include in the snippet. My inclination would be to say, no, I don't think that the collection creator will very often be particularly useful in determining potential relevance, so perhaps we should not include that field in the search result snippet at all.

I don't feel that strongly about it, and it only takes up one line in the snippet, so I don't mind being overruled. If we keep it, maybe let's just go without a label at first, and I'll style it and we can see how it feels with real data.

anarchivist:
OK, now that I'm reading through this again I have some mixed feelings about this. I think my main concern is the case where a component may have an identified creator, and the collection creator also displays. Any thoughts about how to best deal with that case?

ggeisler:
Okay, if we potentially can show both a collection creator and a component creator in the same result snippet, yes, I agree we probably should include at least the collection creator label.

I still wonder if we even need to display creators in a result item. In our mockups we only suggested showing the following in the results items (unless I'm missing something somewhere else):

arclight-mockups-mvp_pdf__page_17_of_23_

So the question for me is still: do we need to show creators in a result item?

anarchivist:
Hm, yeah, now that I think through it, let's try to see what happens when we leave out creators in a result item. There may be cases where it's important, but since we have the data now adding it back is not terribly difficult.

jkeck:
Okay, just want to make sure I'm clear on what the change request is.

It sounds like we should remove the collection creator for the search results display (which this PR adds), but retain the configuration to be passed through the POC Google Request Form implementation.

Does this indicate that we should also remove the creator from the collection search results document as well?

ggeisler:
> Does this indicate that we should also remove the creator from the collection search results document as well?

I'd say Yes, but am happy to let @anarchivist make the call.

anarchivist:
Here are the changes:

* Remove _collection creatorP from the results display
* Keep _collection creator_ indexing code, and pass the collection creator to the the request handler
* Keep the _creator_ in the results display for now

PR #322, opened 2017-05-12
add_self_or_parents_restrictions

gordonleacock:
Took a flying stab at this but it is incomplete so "WIP".

The xpath is probably wrong and I wasn't clear how to do the logic of walking up the parent tree to get all the proper restrictions.

### DIFF
--- a/lib/arclight/solr_ead_indexer_ext.rb
+++ b/lib/arclight/solr_ead_indexer_ext.rb
@@ -14,6 +14,10 @@ module Arclight
 
       add_collection_creator_to_component(node, solr_doc)
 
+      add_self_or_parents_restrictions(node, solr_doc)
+
+      add_self_or_parents_terms(node, solr_doc)
+
       solr_doc
     end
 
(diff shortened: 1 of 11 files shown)


spec/models/concerns/arclight/solr_document_spec.rb
```diff
@@ -86,11 +86,27 @@
     end
   end
 
-  describe '#normalize_date' do
```
drh-stanford: @gordonleacock it looks like the coverage decrease is because you're removing this test for the `normalize_date`

gordonleacock: Thanks Daren - I certainly didn't mean to do that!

lib/arclight/solr_ead_indexer_ext.rb
```diff
@@ -126,5 +130,31 @@ def add_collection_creator_to_component(node, solr_doc)
       creators = node.xpath('//archdesc/did/origination[@label="creator"]/*/text()').map(&:text)
       solr_doc[field_name] = creators - [repository]
     end
+
+    def parent_check_list(node, root_path, element_path, results = [])
```
drh-stanford: This method is similar to the `ancestral_visit` method above and might be reuseable.

PR #323, opened 2017-05-12
normalize collection and component titles to use "Title, Date" throughout the app

### LINKED ISSUE #319: Consistent display of unittitle/unitdate throughout the app
Suggested by @drh-stanford in #305.

The combination of `<unittitle/>` and `<unitdate/>` should be presented consistently throughout the app - component/collection titles, breadcrumbs, etc.

* If there's a `<unittitle/>` with `<unitdate/>`: Join the two with a comma and a space, ensuring no double commas (per #180)
* If there's a `<unittitle/>`, but no `<unitdate/>`: Whatever the `<unittitle/>` is, with no trailing comma
* If there's a `<unittitle/>`, with `<unitdate>n.d.</unitdate>`: Whatever the `<unittitle/>` is + `, n.d.`
* If there's no `<unittitle/>`, with `<unitdate>n.d.</unitdate>`: `n.d.`
* If there's no `<unittitle/>` or `<unitdate/>`: issue a warning and use the component `id`?

*Edit*

* If the `<unitdate/>` has a `type` attribute set to `bulk`, prefix that date with `bulk`

Per @drh-stanford's suggestion we may want to approach this as normalization at index time.

drh-stanford:
This PR fixes #319.

It uses the `Arclight::NormalizeTitle` throughtout the indexer and changes the default title to the `normalized_title_ssm` field.  Titles are now "Title, Date" where the date follows rules in http://www2.archivists.org/standards/DACS/part_I/chapter_2/4_date. Some examples include:

* My Title
* 1994-1999
* My TItle, n.d.
* My Title, c.1999
* My TItle, 1999-2000
* My Title, 1999-2000, bulk 1995-2005

Some notes about this PR:

* It rebuilds the `parent` fields that `solr_ead` generates as _both_ their ids and titles are not normalized as we want them to be. This code is in `solr_ead_indexer_ext.rb` and it replaces much of what was there before.

screen shot 2017-05-16 at 4 11 02 pm
^^^ both ids and titles are normalized ^^^
* It removes the `document.unitdate` in the solr document concern. There's now a `normalized_date` for that information.
* It removes the `normalized_date_range` field and replaces it with a `NormalizeDate` class which follows the DACS rules better.
* I didn't remove `title_ssm` and `unitdate_ssm` from the indexer -- these are the raw text values from those elements and they may be useful at some point.
* We're only working with the `unitdate TYPE` attribute --  I'm not looking at the `unitdate CERTAINTY` attribute for example.

### Collection facet

screen shot 2017-05-17 at 11 13 38 am

### Examples of Contents on Collection show page
screen shot 2017-05-17 at 11 14 43 am

screen shot 2017-05-17 at 11 06 25 am

### DIFF
--- a/lib/arclight/normalized_date.rb
+++ b/lib/arclight/normalized_date.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Arclight
+  ##
+  # A utility class to normalize dates, typically by joining inclusive and bulk dates
+  # e.g., "1990-2000, bulk 1990-1999"
+  # @see http://www2.archivists.org/standards/DACS/part_I/chapter_2/4_date
+  class NormalizedDate
+    # @param [String | Array<String>] `inclusive` from the `unitdate`
+    # @param [String] `bulk` from the `unitdate`
+    # @param [String] `other` from the `unitdate` when type is not specified
+    def initialize(inclusive, bulk = nil, other = nil)
+      if inclusive.is_a? Array # of YYYY-YYYY for ranges
+        @inclusive = YearRange.new(inclusive.include?('/') ? inclusive : inclusive.map { |v| v.tr('-', '/') }).to_s
+      elsif inclusive.present?
+        @inclusive = inclusive.strip
+      end
+      @bulk = bulk.strip if bulk.present?
+      @other = other.strip if other.present?
+    end
+
+    # @return [String] the normalized title/date
+    def to_s
+      normalize
+    end
+
+    private
+
+    attr_reader :inclusive, :bulk, :other
+
+    # @see http://www2.archivists.org/standards/DACS/part_I/chapter_2/4_date for rules
+    def normalize
+      if inclusive.present?
+        result = inclusive.to_s
+        result << ", bulk #{bulk}" if bulk.present?
+      elsif other.present?
+        result = other.to_s
+      else
+        result = nil
+      end
+      return if result.blank?
+      result.strip
+    end
+  end
+end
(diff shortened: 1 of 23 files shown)


lib/arclight/normalized_date.rb
```diff
@@ -0,0 +1,52 @@
...
+      else
+        result = nil
+      end
+      return if result.blank?
+      filter_dates(result)
+    end
+
+    def filter_dates(dates)
+      dates.gsub!('n.d.', 'undated ') # use "undated" instead of "n.d."
+      dates.gsub!('c.', 'circa ') # use "circa" instead of "c."
```
jkeck: I wonder if we should internationalize these strings?  Is it possible that downstream applications may want to customize this string themselves.

Is this potentially problematic due to the lack of rails context during the indexing process?

drh-stanford: Good point. I'm using the terms described in English in DACS 2.4.12 and 2.4.16 -- see http://www2.archivists.org/standards/DACS/part_I/chapter_2/4_date. We don't have a rails context during indexing so I'm not sure how to use the `t()` function here.

anarchivist: Per [#319 (comment)](https://github.com/sul-dlss/arclight/issues/319#issuecomment-301969809), my opinion is that we should not be normalizing them here - we should just accept what exists in the text node already. I think this kind of normalization is out of scope for us.

drh-stanford:
Ok, thanks for the feedback. I took out the normalization for the "c." and "n.d." terms. The PR is ready for review again...

jkeck:
Sorry, didn't notice I got pinged on this. 👍

PR #324, opened 2017-05-13
Add a rake task to destroy all index documents.

jkeck:
We can use this in the demo app to clear out docs before indexing (which will help us w/ the documents w/ the bad IDs).


mejackreed:
I'm 👍  on the code here, but a little 👎  on the idea of having this task in the app. I know people have been bitten before by having this readily handy.

jkeck:
👎 on having the code in which app?  sul-dlss/arclight-demo?

PR #342, opened 2017-05-17
Updates solrconfig/schema for autocomplete.

### LINKED ISSUE #244: Autocomplete/typeahead
**Acceptance criteria**

When a user begins typing query terms into to a search box anywhere in the application:

* the user sees recommendations for their queries below the search box
* the user can select the recommendations either with the keyboard or with the mouse

**Out of scope**

* Field-specific search 
* Type-specific contextual clues (e.g., exclude the "person" labels)

**Mockup**

arclight-mockups-end-user-ui

jkeck:
Closes #244 

This changes the default configuration (that ships w/ Blacklight) from using `FuzzyLookupFactory` to `AnalyzingInfixLookupFactory`.

The result of this is a non-prefixed search.  Using `FuzzyLookupFactory` you must type the beginning of a field.  This means that typing a person's first name (when their name is indexed `Last` `,` `First`) the autocomplete will not return the name.  Also, titles that begin in punctuation would have to have the punctuation typed first in-order to autocomplete.

Talking this over w/ @anarchivist, this does not seem to be the desired behavior for Arclight.

Instead I'm using the `AnalyzingInfixLookupFactory` which allows for non-prefixed autocompleting.

autocomplete

Note: I'm going to break adding the autocomplete to the `Search within this collection` search form into a separate ticket.  This is because the autocomplete feature does not limit the results that are returned based on any context (even in other faceted search results).  I've done a bit of research and will add my findings to that ticket.

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -569,7 +569,12 @@
    <copyField source="*_t" dest="spell"/>
 
    <!-- for suggestions -->
-   <copyField source="*_t" dest="suggest"/>
+   <copyField source="title_ssm" dest="suggest"/>
+   <copyField source="collection_sim" dest="suggest"/>
+   <copyField source="names_ssim" dest="suggest"/>
+   <copyField source="repository_sim" dest="suggest"/>
+   <copyField source="places_ssim" dest="suggest"/>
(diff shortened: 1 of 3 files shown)


drh-stanford:
Resolved a conflict and now waiting on travis...

PR #355, opened 2017-05-19
add slug-based routes for repositories

### LINKED ISSUE #329: Individual repositories should have their own route(s)
**Acceptance criteria**

* Each configured repository should have an easy to remember route based on its slug/ID to the faceted search results for that repository
    * for example `/repositories/hoover` should show me the same content as <br/> `/catalog?f%5Brepository_sim%5D%5B%5D=Hoover+Institution+Archives`
* Each repository linked from the repositories page (`/repositories`) should be linked to its corresponding slug

drh-stanford:
This PR fixes #329. It adds slug based routes like `/repositories/my-slug`. Links for the repositories in the "In person" and "Repositories" view use the slug link. If the slug is not valid, then it throws `ActiveRecord::RecordNotFound`.

## The Repositories index page

Note that the Repository menu is marked "active"

screen shot 2017-05-24 at 10 40 55 am

## The Repository show page

Note the breadcrumbs and both menu items are not marked active.

screen shot 2017-05-24 at 10 41 05 am

## Search results that has hits all from the same repository

Note both menu items are not marked active.

screen shot 2017-05-24 at 10 41 24 am

### DIFF
--- a/app/controllers/arclight/repositories_controller.rb
+++ b/app/controllers/arclight/repositories_controller.rb
@@ -8,6 +8,16 @@ module Arclight
       load_collection_counts
     end
 
+    def show
+      @repository = Arclight::Repository.find_by!(slug: params[:id])
+      search_service = Blacklight.repository_class.new(blacklight_config)
+      @response = search_service.search(
+        q: "level_sim:Collection repository_sim:\"#{@repository.name}\"",
+        rows: 100
+      )
+      @collections = @response.documents
+    end
+
     private
 
     def load_collection_counts
--- a/app/helpers/arclight_helper.rb
+++ b/app/helpers/arclight_helper.rb
@@ -33,12 +33,17 @@ module ArclightHelper
     facets_from_request.find { |f| f.name == 'collection_sim' }.try(:items).try(:count)
   end
 
-  def repositories_active?
-    controller_name == 'repositories'
+  def on_repositories_show?
+    controller_name == 'repositories' && action_name == 'show'
   end
 
+  def on_repositories_index?
+    controller_name == 'repositories' && action_name == 'index'
+  end
+
+  # the Repositories menu item is only active on the Repositories index page
   def repositories_active_class
-    'active' if repositories_active?
+    'active' if on_repositories_index?
   end
 
   def fields_have_content?(document, field_accessor)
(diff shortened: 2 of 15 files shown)


app/controllers/arclight/repositories_controller.rb
```diff
@@ -6,5 +6,12 @@ class RepositoriesController < ApplicationController
     def index
       @repositories = Arclight::Repository.all
     end
+
+    def show
+      repository = Arclight::Repository.find_by!(slug: params[:id])
+      redirect_to search_action_path(
```
mejackreed: The way that I read #329 is that `/repositories/:id` should show the same content as the search, not redirect there. @anarchivist is that correct?

anarchivist: I think that's how I wrote it out; I guess I'm not super opinionated about it, but the intent is in a future work cycle we may want to add a new set of behaviors for the repository show page (e.g., showing a list of "highlighted" collections as shown in in the mockups, etc.)

mejackreed: Ok, my question is also about the url.. should the vanity slug url persist or is a redirect to a catalog faceted one ok? @anarchivist

anarchivist: I think my inclination would be that the slug URL should persist.

drh-stanford: I can work on persisting the vanity slug URL.

drh-stanford:
@mejackreed i've force pushed the branch and it's ready for review again.

app/helpers/arclight_helper.rb
```diff
@@ -34,7 +34,7 @@ def collection_count
   end
 
   def repositories_active?
-    controller_name == 'repositories'
+    controller_name == 'repositories' && action_name == 'index'
```
mejackreed: This seems to turn off the `active` ability of the link here:
screen shot 2017-05-23 at 6 38 45 pm
Should that still be active?

drh-stanford: so i think the behavior is correct but the page is missing a breadcrumb which i think is the source of the confusion here.

mejackreed: Ah. my inclination was that the "Repositories" link in the header would be active to indicate a subpath of "Repositories". This behavior is similar to the way that "Collections" works.

drh-stanford:
@mejackreed I think the interface is less confusing now, but maybe this is an open question as to whether the "Repositories" menu should be marked active on the Repository detail page . I think with the breadcrumbs added it's more clear.

PR #363, opened 2017-05-22
Update styling of inventory contents hierarchy

ggeisler:
This is a just a first-pass at making the contents hierarchy a bit more presentable. There are further styling and layout adjustments to make and other problems to solve (more polished iconography, more visible links to component detail pages) but this seems like a worthwhile step forward.

milton_friedman_papers__1931-2006_-_blacklight

Note that these changes have an untended effect on the component detail, Collection Context section in that the children badge in that context should not show the "View" toggle part of the badge, as shown below. If we merge this PR we can address that separately. I also will do styling of the component detail page in case other unintended styling issues have propagated to that page.

series_ii__membership__1902-1973_-_blacklight

### DIFF
--- a/config/locales/arclight.en.yml
+++ b/config/locales/arclight.en.yml
@@ -7,7 +7,7 @@ en:
       sidebar_section:
         link: 'Open viewer'
     hierarchy:
-      view_more: 'View more'
+      view_all: 'View'
     request:
       container: 'Request container(s)'
     routes:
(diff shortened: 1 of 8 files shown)


config/locales/arclight.en.yml
```diff
@@ -7,6 +7,7 @@ en:
       sidebar_section:
         link: 'Open viewer'
     hierarchy:
+      view_all: 'View'
       view_more: 'View more'
```
mejackreed: Can we remove this now?

ggeisler: Yeah, I left it in case we were using it elsewhere, but you're right, we aren't so I've removed it and updated this PR.

PR #384, opened 2017-05-24
Request all the things

### LINKED ISSUE #287: Display request buttons in components lists
**Acceptance criteria**

In displays where lists of components appear:

* [x] Display a "Request" button for each component with containers
* [x] Clicking a request button should pop up a request form with the right info 

**Questions**

* Should this appear in collection's inventory view as well as the tabular component display (#231)?
* Is this expected to be achieved how do we best select multiple components for requests? Check boxes or something similar?

**Mockup**

arclight-mockups-mvp

mejackreed:
Closes #287 

I recognize that likely there will need to be some design changes follow on for accommodation of this button.

request-everything

### DIFF
--- a/app/controllers/concerns/arclight/field_config_helpers.rb
+++ b/app/controllers/concerns/arclight/field_config_helpers.rb
@@ -15,6 +15,7 @@ module Arclight
         helper_method :before_you_visit_note_present
         helper_method :context_sidebar_visit_note
         helper_method :context_sidebar_containers_request
+        helper_method :item_requestable?
       end
     end
 
@@ -29,6 +30,11 @@ module Arclight
 
     def repository_config_present(_, document)
       document.repository_config.present?
+    end
+
+    def item_requestable?(_, options)
+      document = options[:document]
+      request_config_present('', document) && document.containers.present?
     end
 
     def request_config_present(var, document)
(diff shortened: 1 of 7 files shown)


solr/conf/solrconfig.xml
```diff
@@ -122,7 +122,8 @@
          ref_ssm,
          component_level_isim,
          parent_access_restrict_ssm,
-         parent_access_terms_ssm
+         parent_access_terms_ssm,
+         containers_ssim
```
drh-stanford: `containers_ssim` is already in here on line 106.

mejackreed: Nice catch.. pushed an update.

app/controllers/concerns/arclight/field_config_helpers.rb
```diff
@@ -31,6 +32,12 @@ def repository_config_present(_, document)
       document.repository_config.present?
     end
 
+    def item_requestable?(_, options)
+      document = options[:document]
+      # byebug
```
drh-stanford: can we remove this debugging statement?

mejackreed: 👍 thanks good catch.. pushed.

Issue #375, opened 2017-05-23
Collection show page: indexed names should only show terms at the collection level


anarchivist:
Currently, the collection-level display showing names at all levels in the following locations: 

- Summary → Creator
- Indexed terms →  Names

The collection level display should only show these names that appear at the collection level. 

In this screenshot, "Higgins, L. Raymond" is a name that only appears at the component level. 

In addition, these names should not show up in the facets except at the appropriate level.

**Screenshot**

![screen shot 2017-05-24 at 10 40 02 am](https://cloud.githubusercontent.com/assets/73732/26417489/3515aac8-406e-11e7-8299-0e14286b2b70.png)

PR #385, opened 2017-05-24
issue #375 coll index terms fix

gordonleacock:
Okay, this PR does not address the separate requirements for facets to index all the elements in each finding aid. It merely changes the display of names in the **indexed terms** section of the collection show page so they are limited to the names under the <archdesc> element.

### DIFF
--- a/lib/arclight/custom_document.rb
+++ b/lib/arclight/custom_document.rb
@@ -60,20 +60,24 @@ module Arclight
         { name: 'creators', value: creators, index_as: :symbol },
         { name: 'has_online_content', value: online_content?, index_as: :symbol },
         { name: 'repository', value: repository_as_configured(repository), index_as: :displayable },
-        { name: 'repository', value: repository_as_configured(repository), index_as: :facetable }
+        { name: 'repository', value: repository_as_configured(repository), index_as: :facetable },
+        { name: 'names_coll', value: names_coll, index_as: :symbol }
       ]
     end
-    # rubocop: enable Metrics/MethodLength
 
     def names
       [corpname, famname, name, persname].flatten.compact.uniq - repository
+    end
+
+    def names_coll
+      names_array(%w[corpname famname name persname], parent: 'archdesc')
     end
 
     def creators
       [creator_persname, creator_corpname, creator_famname].flatten.compact.uniq - repository
     end
 
-    # Combine subjets into one group from:
+    # Combine subjets into one group from
     #  <controlaccess/><subject></subject>
     #  <controlaccess/><function></function>
     #  <controlaccess/><genreform></genreform>
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,6 +33,10 @@ Metrics/BlockLength:
     - 'tasks/**/*'
     - 'lib/arclight/custom_document.rb'
 
+Metrics/MethodLength:
+  Exclude:
+    - 'lib/arclight/custom_document.rb' # arclight_field_definitions too long
+
 Performance/RegexpMatch:
   Enabled: false
 
--- a/lib/arclight/shared_indexing_behavior.rb
+++ b/lib/arclight/shared_indexing_behavior.rb
@@ -20,6 +20,12 @@ module Arclight
       clean_facets_array(subjects.flatten.map(&:text))
     end
 
+    def names_array(elements, parent:)
+      xpath_elements = elements.map { |el| "local-name()='#{el}'" }.join(' or ')
+      names = search("//#{parent}/controlaccess/*[#{xpath_elements}]").to_a
+      clean_facets_array(names.flatten.map(&:text))
+    end
+
     # Return a cleaned array of facets without marc subfields
     #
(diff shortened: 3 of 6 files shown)


lib/arclight/custom_document.rb
```diff
@@ -21,6 +21,7 @@ class CustomDocument < SolrEad::Document
       t.function(path: 'archdesc/controlaccess/function/text()', index_as: %i[displayable facetable])
       t.occupation(path: 'archdesc/controlaccess/occupation/text()', index_as: %i[displayable facetable])
       t.places(path: 'archdesc/controlaccess/geogname/text()', index_as: %i[displayable facetable symbol])
+      t.places_coll(path: "archdesc[@level='collection']/controlaccess/geogname/text()", index_as: %i[displayable facetable symbol])
```
mejackreed: I see that we are using `places_coll_ssim`. Do we still then need `displayable` and `facetable`?

gordonleacock: Hi @mejackreed - Yes, this is being used in a facet. I can try removing displayable.

@anarchivist - The purpose of this PR is to fix the collection level facets to not include those lower level heading.

gordonleacock: @mejackreed - Actually it works if I just leave symbol. Will soon push that up. Thanks.

lib/arclight/custom_document.rb
```diff
@@ -21,6 +21,7 @@ class CustomDocument < SolrEad::Document
       t.function(path: 'archdesc/controlaccess/function/text()', index_as: %i[displayable facetable])
       t.occupation(path: 'archdesc/controlaccess/occupation/text()', index_as: %i[displayable facetable])
       t.places(path: 'archdesc/controlaccess/geogname/text()', index_as: %i[displayable facetable symbol])
+      t.places_coll(path: "archdesc[@level='collection']/controlaccess/geogname/text()", index_as: %i[symbol])
```
anarchivist: Per Slack discussion, I'm not sure this is necessary.

lib/arclight/custom_document.rb
```diff
@@ -63,12 +64,22 @@ def arclight_field_definitions
...
+        { name: 'access_subjects_coll', value: access_subjects_coll, index_as: :symbol }
+      ]
+    end
 
     def names
       [corpname, famname, name, persname].flatten.compact.uniq - repository
     end
 
+    def names_coll
+      names_array(%w[corpname famname name persname], parent: "archdesc[@level='collection']")
```
anarchivist: We should remove the `[@level='collection']` part of the XPath - when I said "collection level" I meant at the top level of description, i.e. the `<archdesc/>`.

lib/arclight/custom_document.rb
```diff
@@ -81,5 +92,9 @@ def creators
     def access_subjects
       subjects_array(%w[subject function occupation genreform], parent: 'archdesc')
     end
+
+    def access_subjects_coll
+      subjects_array(%w[subject function occupation genreform], parent: "archdesc[@level='collection']")
```
anarchivist: Per Slack discussion, I'm not sure this is necessary. `access_subjects` gets us what we need already.

lib/arclight/custom_document.rb
```diff
@@ -63,21 +63,25 @@ def arclight_field_definitions
...
 
+    def names_coll
+      names_array(%w[corpname famname name persname], parent: 'archdesc')
+    end
+
     def creators
       [creator_persname, creator_corpname, creator_famname].flatten.compact.uniq - repository
     end
 
-    # Combine subjets into one group from:
```
mejackreed: Is this comment deletion intentional?

lib/arclight/custom_document.rb
```diff
@@ -31,7 +31,7 @@ class CustomDocument < SolrEad::Document
     def to_solr(solr_doc = {})
       super
       solr_doc['id'] = Arclight::NormalizedId.new(eadid.first).to_s
-      arclight_field_definitions.each do |field|
+      [*arclight_field_definitions, *arclight_field_definitions_colls].each do |field|
```
jkeck: Is there any particular reason that we're splitting these field definitions into two separate arrays?  I'm not really following what about the `_colls` makes this array different than `arclight_field_definitions`.

lib/arclight/custom_document.rb
gordonleacock: Nope, I'll fix.

lib/arclight/custom_document.rb
gordonleacock: Rubo cop was complaining about the size so I split it into two.

gordonleacock: Glad to do something else if you have a suggestion!

jkeck: Hmm, I feel like this might be the point where we're adding unnecessary cognitive load just to appease the cops.  I would be 👍 to excluding this file from the LineLength cop `.rubocop.yml` (with a comment explaining why).

PR #390, opened 2017-05-24
Update layout and styling of Repository show page

ggeisler:
* Updated the button (now a link styled as a button) on the Repositories index page
* Added a "View all collections" link to the Repository show page (not so useful for us now, but as soon as a repo has more than a handful of collections it will be)
* Updated the layout and styling to look a bit nicer at various viewport sizes

Note that the "View all of our collections" link needs to be updated to actually go somewhere. I don't know the proper route so I just put in a placeholder (with a TODO in the code). The link should go to search results scoped to a) the current repository and b) Level = 'collection'

I can write up a new ticket for that link TODO, or someone could push a commit to this PR. Also, I'm not sure if the two links I added should have tests; if so, probably better for someone else to do that, again either as a new commit to this branch or we can write a new ticket.

### Example Repository show page
blacklight 2

### DIFF
--- a/app/assets/stylesheets/arclight/application.scss
+++ b/app/assets/stylesheets/arclight/application.scss
@@ -5,6 +5,7 @@
 @import 'modules/hierarchy';
 @import 'modules/layout';
 @import 'modules/mastheads';
+@import 'modules/repositories.scss';
 @import 'modules/repository_card.scss';
 @import 'modules/search_results';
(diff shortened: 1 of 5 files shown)


camillevilla:
@ggeisler @jkeck suggests creating a ticket for that TODO.

ggeisler:
Thanks @camillevilla! 

New ticket is #394.

PR #392, opened 2017-05-24
Add a compact view for Blacklight search results.

### LINKED ISSUE #247: Compact view of search results
**Acceptance criteria**

On a search results page:

* Display a dropdown menu in the same section as the sort/pagination controls to switch between "Full View" (current default) and "Compact View"
* When a user switches to compact view:
* Display a three-column display, which displays the collection or component's title/date in the left column, and (for components only) breadcrumb links for all of the parents in the middle column, and an online content label (if appropriate)

**Mockup**

arclight-mockups-mvp

jkeck:
Closes #247

@ggeisler do you have any thoughts on an SVG icon for this?  Also, we're going w/ the default Blacklight behavior for view type switching (which is not a drop-down), assuming that a downstream application can make that customization if they desire.  Please let us know if we should spend time on customizing the view picker to be a drop-down.

## Example Search Result
compact-view

### DIFF
--- a/app/assets/stylesheets/arclight/modules/search_results.scss
+++ b/app/assets/stylesheets/arclight/modules/search_results.scss
@@ -54,3 +54,17 @@
 .catalog_startOverLink {
   margin-bottom: ($spacer / 2);
 }
+
+// Compact Search Results
+.documents-compact {
+  margin-bottom: $spacer;
+
+  .document {
+    border-bottom: $result-item-border-styling;
+    padding: $spacer 0;
+
+    &:first-child {
+      border-top: $result-item-border-styling;
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


ggeisler:
@jkeck Looks good. (Although I'm concerned about the user understanding the difference between the two columns; JV and I never came up with a solution to this, though I wonder if column headings (e.g., "Item" & "Collection location") might be enough -- but that is probably not a concern of this PR).

I'm okay with your plan for the view switching, though it will be really helpful if we can find an icon that clearly represents "compact" if we're relying on the icons for switching. I don't have any ideas off the top of my head.

jkeck:
Created #397 for adding the icon.

PR #398, opened 2017-05-25
add indexing of component subjects

gordonleacock:
Makes a small change in lib/arclight/custom_document.rb to get all the terms instead of those under <archdesc> only.

### DIFF
--- a/lib/arclight/custom_component.rb
+++ b/lib/arclight/custom_component.rb
@@ -18,7 +18,8 @@ module Arclight
       t.ref_(path: '/c/@id', index_as: %i[displayable])
 
       # facets
-      t.creator(path: "c/did/origination[@label='creator']/*/text()", index_as: %i[displayable facetable])
+      t.creator(path: "c/did/origination[@label='creator']/*/text()", index_as: %i[displayable facetable symbol])
+      t.places(path: 'c/controlaccess/geogname/text()', index_as: %i[displayable facetable symbol])
 
       add_unitid(t, 'c/')
       add_extent(t, 'c/')
@@ -50,7 +51,7 @@ module Arclight
     def component_field_definitions
       [
         { name: 'level', value: formatted_level, index_as: :facetable },
-        { name: 'access_subjects', value: access_subjects, index_as: :facetable },
+        { name: 'access_subjects', value: access_subjects, index_as: :symbol },
         { name: 'containers', value: containers, index_as: :symbol },
         { name: 'has_online_content', value: online_content?, index_as: :symbol }
       ]
--- a/spec/features/indexing_custom_document_spec.rb
+++ b/spec/features/indexing_custom_document_spec.rb
@@ -134,6 +134,10 @@ RSpec.describe 'Indexing Custom Document', type: :feature do
       expect(doc['normalized_date_ssm'].first).to eq '1894-1992'
     end
 
+    it '#names_coll' do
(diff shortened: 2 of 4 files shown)


gordonleacock:
Having to back off and start over with this one. The broader issue is that all the facets need to index their respective elements at the component levels as well as at the collection level.

spec/features/indexing_custom_document_spec.rb
```diff
@@ -74,7 +74,6 @@
 
     it '#level' do
       expect(doc['level_ssm'].first).to eq 'collection'
-      expect(doc['level_sim'].first).to eq 'Collection'
```
drh-stanford: I don't see why this test is being removed.

spec/features/indexing_custom_document_spec.rb
```diff
@@ -92,14 +91,12 @@
 
       it 'matches EAD data' do
         expect(doc['repository_ssm'].first).to eq '1118 Badger Vine Special Collections'
-        expect(doc['repository_sim'].first).to eq '1118 Badger Vine Special Collections'
```
drh-stanford: Same here. I don't see why this test is being removed.

spec/features/indexing_custom_document_spec.rb
```diff
@@ -92,14 +91,12 @@
...
       end
 
       context 'with REPOSITORY_ID' do
         it 'matches Repository configuration' do
           ENV['REPOSITORY_ID'] = 'nlm'
           expect(doc['repository_ssm'].first).to eq 'National Library of Medicine. History of Medicine Division'
-          expect(doc['repository_sim'].first).to eq 'National Library of Medicine. History of Medicine Division'
```
drh-stanford: And here. I don't see why this test is being removed.

lib/arclight/custom_component.rb
```diff
@@ -46,9 +47,10 @@ def to_solr(solr_doc = {})
     def component_field_definitions
       [
         { name: 'level', value: formatted_level, index_as: :facetable },
-        { name: 'access_subjects', value: access_subjects, index_as: :facetable },
+        { name: 'access_subjects', value: access_subjects, index_as: :symbol },
         { name: 'containers', value: containers, index_as: :symbol },
-        { name: 'has_online_content', value: online_content?, index_as: :symbol }
+        { name: 'has_online_content', value: online_content?, index_as: :symbol },
+        { name: 'places', value: places, index_as: :facetable }
```
drh-stanford: I think this is redundant due to the terminology indexing it already. When you add a term to the terminology, it's automagically indexed. So since line 22 has `facetable` as one of the `index_as` parameters, I believe this is redundant.

spec/features/indexing_custom_document_spec.rb
gordonleacock: No problem, I though that a duplicate line had been created in error, I see now that they were different. I'll fix. Thanks!

spec/features/indexing_custom_document_spec.rb
gordonleacock: Ditto.

lib/arclight/custom_component.rb
gordonleacock: Hi Darren, I assume you are talking about the line with 'places' which I've now removed. Thanks.

PR #404, opened 2017-05-25
add sorting by title, date, and creator

### LINKED ISSUE #367: Sorting
**Acceptance criteria**

From a set of search results, allow a user to sort by the following criteria:

* relevance (default)
* date 
* creator
* title

drh-stanford:
This PR fixes #367. It creates new `title_sort`, `creator_sort`, and `date_sort` fields. It also changes the hiearchy sort parameter to `sort_ii asc` which preserves the order of components in the EAD.

Note that the default sorting before this PR was wonky (it was only using `title_filing_si` which is the `titleproper` field and not present for components 😞 ). Therefore, this PR changes some of the failing tests to match the result order with the correct ordering by `title_sort`.

Note that creator is multi-valued so `copyField` doesn't work for it (title and date are also multivalued but in practice they're single-valued -- marked this with a TODO). Also adds a missing test for the highlighting search builder behavior.

screen shot 2017-05-25 at 4 15 16 pm

### DIFF
--- a/spec/models/concerns/arclight/search_behavior_spec.rb
+++ b/spec/models/concerns/arclight/search_behavior_spec.rb
@@ -44,4 +44,18 @@ describe Arclight::SearchBehavior do
       end
     end
   end
+  describe '#add_highlighting' do
+    context 'when in hierarchy view' do
+      let(:user_params) { { view: 'hierarchy' } }
+
+      it 'disables highlighting' do
+        expect(search_builder_instance.add_highlighting(solr_params)).to include('hl' => false)
+      end
+    end
+    context 'when not in hierarchy view' do
+      it 'enables highlighting' do
+        expect(search_builder_instance.add_highlighting(solr_params)).to include('hl' => true)
+      end
+    end
+  end
 end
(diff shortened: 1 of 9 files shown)


spec/models/concerns/arclight/search_behavior_spec.rb
```diff
@@ -30,4 +30,32 @@
...
+        expect(search_builder_instance.add_hierarchy_sort(solr_params)).to include(sort: 'sort_ii asc')
+      end
+    end
+    context 'when not in hierarchy view' do
+      it 'does not affect sort param' do
+        expect(search_builder_instance.add_hierarchy_sort(solr_params)).to eq({})
+      end
+    end
+  end
+  describe '#add_highlighting' do
```
camillevilla: just reviewed this one with @jkeck. @drh-stanford, could you separate these highlighting tests into a different commit?

drh-stanford: thanks for the feedback. i've put the highlight spec changes into a different commit in this PR.

Issue #381, opened 2017-05-24
Add documentation to the wiki on how to update an ArcLight app


anarchivist:
We need to provide some documentation about how someone can update an existing ArcLight application.

Context: Folks at the Bentley have tested ArcLight and ran into #361; we should now provide directions for them to follow about how to update their app make sure the fixes show up.

drh-stanford:
See https://github.com/sul-dlss/arclight/wiki/Upgrading-your-ArcLight-application

PR #413, opened 2017-05-26
[PROPOSAL] Add an Update generator that does the install steps in a non-forceful way

jkeck:
Wondering if this could be useful in thinking about #381.

For our demo app we simply run `rails g arclight:install` to update it each time (since we don't customize anything really).  The things that we generate during our installation, that commonly need to be updated are the solr and catalog controller configurations.  Our installer forcefully updates both of these (for necessary reasons).

This PR introduces an update generator (`rails g arclight:update`) that subclasses the install generator.  This has the benefit of easily allowing downstream applications to get additional generated code (e.g. added initializers, insertions into downstream classes, etc) while allowing us to override certain parts of the install process to be non-forceful.

In the gif below you can see that my local changes to the catalog controller are preserved as I was given the option to overwrite my changes (I choose to view a diff so I can see what the upstream changes are before deciding not to accept them).

al-update

### DIFF
--- a/lib/generators/arclight/update_generator.rb
+++ b/lib/generators/arclight/update_generator.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'generators/arclight/install_generator'
+
+module Arclight
+  ##
+  # Arclight Update generator. This subclasses the Install generator, so this is
+  # intended to override behavior in the install generator that can allow the
+  # downstream application to choose if they want to take our changes or not and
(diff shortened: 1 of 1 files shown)


mejackreed:
It might in some ways, but in other ways (like here: https://github.com/sul-dlss/arclight/blob/master/lib/generators/arclight/install_generator.rb#L16-L20) would we have to check if that thing is already injected?

Issue #378, opened 2017-05-23
Add view toggle for truncated text in contents hierarchy


ggeisler:
We're currently truncating the scope note/other metadata shown for a component, as per the mockups. However, we don't give the user a way expand the truncated text to see the full metadata for the component.

I think we just need to add a simple toggle after the truncated text to provide this functionality. Something like this:

![milton_friedman_papers__1931-2006_-_blacklight](https://cloud.githubusercontent.com/assets/101482/26380917/2ae3ffe8-3fd6-11e7-93d0-ab17f0eefdb1.png)

----

![milton_friedman_papers__1931-2006_-_blacklight 2](https://cloud.githubusercontent.com/assets/101482/26380914/270d9d5c-3fd6-11e7-9c2d-3d834753a2f4.png)

jkeck:
There are several places where this content is truncated.  We'll probably want to think about how we'll implement this type of feature holistically throughout the app.  (e.g. https://github.com/sul-dlss/arclight/blob/master/app/views/arclight/repositories/show.html.erb needs to have its abstract/scope truncated in the same way)

PR #415, opened 2017-05-26
Add section headings to truncated component metadata

### LINKED ISSUE #379: Add section heading(s) for components in contents hierarchy
@anarchivist I don't think we omitted this intentionally, but please correct this ticket if it was discussed and I just missed or forgot the decision.

Currently we're showing a truncated version of metadata for a component (usually just the Scope and Contents note in our fixture/test data) in the hierarchy view. in the mockups (e.g, page 6 of the MVP version) we specified that the truncated text be prefaced with a section heading. I think this is worth adding, as it helps to define each component a bit better. Also, when collections have more and/or longer metadata than our test data, the section headings will be more informative.

So adding them something like this:

milton_friedman_papers__1931-2006_-_blacklight 3

Section headings should be `<h4>`s using this styling (though I can do this part):

``` css
color: #636c72;
font-size: $font-size-xs;
font-weight: normal;
text-transform: uppercase;
```

camillevilla:
Closes #379 

This adds "Scope and Contents" to components in the hierarchy view. I added a feature test for the truncated component metadata as well. The test assumes that truncation length can change but the metadata should end in ellipses. 

Feedback would be much appreciated! 

## Screenshot

scope_and_contents_heading

### DIFF
--- a/app/assets/stylesheets/arclight/modules/hierarchy.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy.scss
@@ -36,6 +36,13 @@
     line-height: 1.25;
   }
 
+  .al-document-abstract-or-scope-heading {
+    color: #636c72;
+    font-size: $font-size-xs;
+    font-weight: normal;
+    text-transform: uppercase;
+  }
+
   // Series level
   .blacklight-series .document-title-heading,
   .blacklight-file .document-title-heading,
--- a/app/assets/stylesheets/arclight/modules/hierarchy.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy.scss
@@ -36,8 +36,8 @@
     line-height: 1.25;
   }
 
-  .al-document-abstract-or-scope-heading {
-    color: #636c72;
+  .al-hierarchy-sub-heading {
+    color: $gray-light;
     font-size: $font-size-xs;
     font-weight: normal;
     text-transform: uppercase;
--- a/app/views/catalog/_index_hierarchy_default.html.erb
+++ b/app/views/catalog/_index_hierarchy_default.html.erb
@@ -1,4 +1,5 @@
 <%= content_tag('div', class: 'al-document-abstract-or-scope', title: document.abstract_or_scope) do %>
+  <%= content_tag(:h4,'Scope and Contents', class: 'al-document-abstract-or-scope-heading') %>
   <%= truncate(document.abstract_or_scope, length: 175) %>
 <% end if document.abstract_or_scope %>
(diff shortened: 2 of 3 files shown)


camillevilla:
Hmm. Just realized that my test will break once #378 gets implemented. Stay tuned for a more content agnostic test.

app/assets/stylesheets/arclight/modules/hierarchy.scss
```diff
@@ -36,6 +36,13 @@
     line-height: 1.25;
   }
 
+  .al-document-abstract-or-scope-heading {
+    color: #636c72;
```
jkeck: Can we move this color to a named variable?

app/views/catalog/_index_hierarchy_default.html.erb
```diff
@@ -1,4 +1,5 @@
 <%= content_tag('div', class: 'al-document-abstract-or-scope', title: document.abstract_or_scope) do %>
+  <%= content_tag(:h4,'Scope and Contents', class: 'al-document-abstract-or-scope-heading') %>
```
jkeck: Can we internationalize this string?  This will allow downstream applications to more easily control this label in the UI.

app/assets/stylesheets/arclight/modules/hierarchy.scss
```diff
@@ -36,6 +36,13 @@
     line-height: 1.25;
   }
 
+  .al-document-abstract-or-scope-heading {
```
jkeck: What would you think about making this class a bit more generic?  Perhaps something that describes what it is, but is also generic enough that it could be used in other contexts.

app/assets/stylesheets/arclight/modules/hierarchy.scss
ggeisler: This is probably my doing, since I gave that example in the ticket. I think `$gray-light;` might be it.

app/assets/stylesheets/arclight/modules/hierarchy.scss
camillevilla: How about `.al-hierarchy-sub-heading`? I found a `.al-show-sub-heading` in _collection_overview.html.erb that seems similar. It's specific to the context, but not the metadata type.

app/assets/stylesheets/arclight/modules/hierarchy.scss
camillevilla: That's the one. Thanks @ggeisler!

PR #420, opened 2017-05-30
Add tasks to index EAD from URL

mark-cooper:
https://github.com/sul-dlss/arclight/issues/418

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,9 +28,9 @@ Metrics/ClassLength:
 
 Metrics/BlockLength:
   Exclude:
+    - !ruby/regexp /\.rake$/
     - 'arclight.gemspec'
     - 'spec/**/*'
-    - 'tasks/**/*'
     - 'lib/arclight/custom_document.rb'
 
 Metrics/MethodLength:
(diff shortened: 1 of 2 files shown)


jkeck:
@mark-cooper looks like there are a few violations of our ruby style guide.

I would be more than happy to work on correcting these with you if you're unfamiliar w/ rubocop.

mark-cooper:
@jkeck I pushed an update. There's still 1 error that I'm not sure what to do with:

```
Offenses:

lib/tasks/index.rake:18:1: C: Metrics/BlockLength: Block has too many lines. [52/25]
namespace :arclight do ...
^^^^^^^^^^^^^^^^^^^^^^

86 files inspected, 1 offense detected
```

Do you have some guidance for how this should be organized? Thanks!

jkeck:
@mark-cooper can you toss it into the `.rubocop.yml`?  We already have some [exemptions](https://github.com/sul-dlss/arclight/blob/master/.rubocop.yml#L29-L34).  It seems like we intended to exclude all rake tasks w/ `tasks/**/*`.  Perhaps we should change that to be just `*.rake` if that works, otherwise adding the `lib/tasks/**/*` would be 👍 too.

anarchivist:
Thanks @mark-cooper! 👏

PR #423, opened 2017-05-30
Major README updates and adding team, resources links

### LINKED ISSUE #411: Update README to acknowledge team and provide basic links to other resources
* We should acknowledge the state of ArcLight at the end of the work cycle
* List contributors, especially Gordon and University of Michigan
* Point to wiki(s), Google Group, etc., Blacklight, etc.

drh-stanford:
This PR fixes #411.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,20 @@
 
 # Arclight
 
-Welcome to your new gem! In this directory, you'll find the files you need to be able to package up your Ruby library into a gem. Put your Ruby code in the file `lib/arclight`. To experiment with that code, run `bin/console` for an interactive prompt.
+A Rails engine supporting discovery of archival materials.
 
-TODO: Delete this and the text above, and describe your gem
+During April - June 2017, ArcLight underwent its initial development as a Minimally Viable Product (MVP). The MVP is targeted for content that is described in the [Encoded Archival Description](http://eadiva.com/2/) (EAD) format. Future development is in the planning phase with no definitive timeline as yet.
+
+## Requirements
+
+* [Ruby](https://www.ruby-lang.org/en/) 2.2 or later
+* [Rails](http://rubyonrails.org) 5.0 or later
 
 ## Installation
 
-Add this line to your application's Gemfile:
+[Installing ArcLight](https://github.com/sul-dlss/arclight/wiki/Creating,-installing,-and-running-your-ArcLight-application) is straightforward in a Rails environment.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,15 +33,55 @@ Or install it yourself as:
...
+  * [Jennifer Vine](https://github.com/jvine)
+  * [Jessie Keck](https://github.com/jkeck)
+  * [Mark Matienzo](https://github.com/anarchivist)
+* University of Michagan
+  * [Gordon Leacock](https://github.com/gordonleacock)
 
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `bundle exec rake` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
+To install this gem onto your local machine, run
```
camillevilla: Hi @drh-stanford! @jkeck and I had a question about this one: Is this step necessary for development?

drh-stanford: Yeah, I've never done a `bundle exec rake install` before, but it was in there so I left it. Should I take it out?

jkeck: @drh-stanford I think it might be unnecessary in our rails engine environment. It's going to create a packaged version of the gem and install that.  In our engine world, we instead reference the root of the engine in the `Gemfile` (using `path:`), so I'm not sure we make use of any packaged/installed gem for this.

Unless anybody else uses this as part of their development workflow, I think we may want to remove it.

drh-stanford: Ok, thanks for the feedback. I've forced pushed up changes to that omit that part of the Development section, and points users to our Wiki for the full installation instructions.

drh-stanford:
@anarchivist do you want to 👍 / 👎 this PR?

README.md
```diff
@@ -4,13 +4,20 @@
 
 # Arclight
 
-Welcome to your new gem! In this directory, you'll find the files you need to be able to package up your Ruby library into a gem. Put your Ruby code in the file `lib/arclight`. To experiment with that code, run `bin/console` for an interactive prompt.
+A Rails engine supporting discovery of archival materials described in the [Encoded Archival Description](http://eadiva.com/2/) (EAD) format.
```
anarchivist: I'd be less specific here but add a note (perhaps in the MVP description) that initial development targeted ingest of EAD 2002.

drh-stanford:
Ok, I've made that update. The PR is ready for review again.

PR #425, opened 2017-05-30
add Download dropdown to Collection show page

### LINKED ISSUE #371: Single page view dropdown: display links to download PDF and EAD
Refs #150. 

**Acceptance criteria** 

Given a EAD file used for indexing:

* Provide a configurable pathway to add a link to enable downloading an EAD version of the finding aid
* Provide a configurable pathway to add a link to enable downloading a PDF version of the finding aid

**Discussion**

* @ggeisler was suggesting that we use pre-created PDF files for this which is fine by me. Any suggestions about the pattern we should follow for this in terms of where to store them? 

**Mockup**

arclight-mockups-mvp

drh-stanford:
This PR fixes #371 by adding a dropdown menu on the Collection show page for downloading any PDF or EAD (.xml) files found in the `config/downloads.yml` metadata. The dropdown only has entries for downloads that are present (i.e., if there's only a PDF configured then the dropdown is just the PDF). If there's no downloads available, it omits the dropdown entirely.

screen shot 2017-05-30 at 2 21 26 pm

The `config/downloads.yml` for the above screenshot would look like this: note both `:href` and `:size` are required:

```yml
M0198:
  pdf:
    href: 'http://example.com/M0198.pdf'
    size: '5 MB'
  ead:
    href: '/documents/M0198.xml'
    size: '30 KB'
```

### DIFF
--- a/app/helpers/arclight_helper.rb
+++ b/app/helpers/arclight_helper.rb
@@ -42,6 +42,17 @@ module ArclightHelper
     facets_from_request.find { |f| f.name == 'collection_sim' }.try(:items).try(:count)
   end
 
+  ##
+  # Looks for `document.unitid` in the downloads configuration
+  # @param [SolrDocument] `document`
+  # @param [Hash] `config` metadata for downloadable files
+  # @return [Hash] with `:href` and `:size` keys
+  def collection_downloads(document, config = load_download_config)
+    config = config[document.unitid] if config.present?
+    return {} if config.blank?
+    parse_collection_downloads(config)
+  end
+
   def on_repositories_show?
     controller_name == 'repositories' && action_name == 'show'
   end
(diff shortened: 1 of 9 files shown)


mejackreed:
On review of this PR and discussion in Slack, I'm curious about the best way forward to provide links to things at the collection level. Part of my initial concern stemmed from using the presence of a file in a specific directory as the way to determine whether or not to show and provide download access to the link. What about remote files? Is this something people actually do?

Could we provide this in the ead instead? If we did that where would it sit? Is that even feasible?

@drh-stanford mentioned having a `downloads.yml` file that could list all of the download information for each collection.

Open to other alternatives (and of course the original approach) but I wanted to at least open the discussion to what actually people could/would do when implementing.

jkeck:
I have concerns about the feasibility of getting this data from the EAD itself.  Unless there is an existing way to model this kind of thing that is "widely" used amongst archives, it seems like `downloads.yml` may be our best option.  This gives downstream appls the flexibility in how they want to manage this data.   If they want it in an S3 bucket, so be it if they want it in their public directory, 👍, if they want it on GitHub, 🆒.

Also, thinking forward to future developments, this might be the sort of thing we want configurable in the admin dashboard, so configuring in a yaml file would be similar to other places where we've followed that same pattern.

drh-stanford:
Ok, I'll take the PR down the `downloads.yml` route. The data that's needed is fairly simple:

```yml
unitid:
  pdf:
    url: http://example.com/unitid.pdf
    size: 123
  ead: ... 
```

anarchivist:
The `downloads.yml` approach seems to be the best option. There is no reliable way to get this information from the EAD itself.

drh-stanford:
Ok, this PR is ready for review. I've updated the top comment with the `downloads.yml` context.

jkeck:
@drh-stanford this looks great to me.

Quick question (which is potentially for @anarchivist): Should we provide any pointer to a real EAD/PDF in either one of our fixture objects, or one of the fixtures in our demo site?

Also, should we document this somewhere? (which doesn't necessarily need to hold up this PR, just a general question).

anarchivist:
> Should we provide any pointer to a real EAD/PDF in either one of our fixture objects, or one of the fixtures in our demo site?

Yep. If I get some pointers I can add a PR to the demo app repo.

> Also, should we document this somewhere? 

Yes, we absolutely should. 😸

jkeck:
Okay, so it sounds like we don't need to hold up this PR then. 👍

Issue #416, opened 2017-05-26
Rework embedded digital object display for collections and components


anarchivist:
This is a substantial reworking of the behavior around digital objects per discussion on #332 and #277.

Refs #242, #237 and #278. 

**Acceptance criteria**

* Remove the "Online content" sidebar block for both collections and components. 
* Add a new "Online content" tab for both collection and component views. 
* For collection-level views:
     * For each digital object at the collection level: 
          * If the object is embeddable, display an embedded viewer for that object
          * If the object is not embeddable, display a link to the digital object 
              * Add a link to the URI identified in the `href` link with the text (in order of preference): as
                   * the `<dao/>`'s `title` attribute OR its `<daodesc/>` 
                   * the `<dao/>`'s  `<daodesc/>` 
                   * the `<dao/>`'s  `href` attribute
     * ~If there are no objects at the collection level, but there are digital objects at the component level, display list of components with digital objects~ (see #416)
     * If there are no collection or component level digital objects:
          * The tab should be inactive and read "No online content"
* For component-level views:
     * For each digital object associated with a component: 
          * If the object is embeddable, display an embedded viewer for that object
          * If the object is not embeddable, display a link to the digital object 
              * Add a link to the URI identified in the `href` link with the text (in order of preference): as
                   * the `<dao/>`'s `title` attribute OR its `<daodesc/>` 
                   * the `<dao/>`'s  `<daodesc/>` 
                   * the `<dao/>`'s  `href` attribute   
     * If there are no digital objects at that level:
          * The tab should be inactive and read "No online content"

**Screenshots**
### Collection-level online content
![viewer](https://cloud.githubusercontent.com/assets/101482/26180086/21176fa4-3b1c-11e7-9aa8-e5071db09c6f.png)

###  Component-level online content
![all-online-content](https://cloud.githubusercontent.com/assets/101482/26180091/2d73b488-3b1c-11e7-851c-929e68d2c344.png)

jkeck:
A few questions looking over this ticket:

1. This ticket assumes that there is an `Overview` and `Contents` tab on the component view.  Currently we have two sections (not in a tabbed UI) that is the Metadata and the Collection Context.  Is part of the acceptance criteria for this ticket to move those sections into tabs for the Component display page?
1. Do we have a fixture w/ an oEmbeddable digital object at the collection level?
[...]

PR #427, opened 2017-05-31
Update fixture for #416

anarchivist:


jkeck:
We can build on-top of this for our PR.  👍 And it's an item from one of my favorite collections!

Issue #293, opened 2017-05-08
Document or provide way to check indexer's expectations for EAD


anarchivist:
Suggested by @cazzerson. We should give people guidelines about the expectations that ArcLight or the indexer puts on EAD finding aids.

Examples:

* All components must have an `id` attribute
* All components must have at least `<unittitle/>` or `<unitdate/>`

Potential forms:

* documentation
* Schematron file
* Helpful warnings or errors during indexing

mejackreed:
Is it acceptable to raise an error on indexing with a helpful error message?

anarchivist:
Yes, that would be acceptable.

mejackreed:
Some more documentation added: https://github.com/sul-dlss/arclight/wiki/Indexing-EAD-in-ArcLight/_compare/b01984298bcd6a5bb48deec2b88b607d00890aa5...40bb48f1975ec8dac65d7b27197bed61d938ef0c

drh-stanford:
this is merged now

PR #435, opened 2017-06-01
raises exceptions in normalization process when expected things are n…

mejackreed:
…ot found

Partially addresses #293. Though this may be controversial. I think documentation will also help. Additional enhancements could be made to the indexing code to print out the component in question?

Thoughts?

### DIFF
--- a/spec/lib/arclight/normalized_id_spec.rb
+++ b/spec/lib/arclight/normalized_id_spec.rb
@@ -24,8 +24,11 @@ RSpec.describe Arclight::NormalizedId do
   context 'when the id is nil' do
     let(:id) { nil }
 
-    it 'returns an empty string' do
-      expect(normalized_id).to eq ''
+    it do
+      expect { normalized_id }.to raise_error(
+        Arclight::Exceptions::IDNotFound,
+        'id must be present for all documents and components'
+      )
     end
   end
 end
--- a/spec/lib/arclight/normalized_title_spec.rb
+++ b/spec/lib/arclight/normalized_title_spec.rb
@@ -3,7 +3,7 @@
 require 'spec_helper'
 
 RSpec.describe Arclight::NormalizedTitle do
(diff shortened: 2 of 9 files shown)


spec/lib/arclight/normalized_id_spec.rb
```diff
@@ -24,8 +24,6 @@
   context 'when the id is nil' do
     let(:id) { nil }
 
-    it 'returns an empty string' do
-      expect(normalized_id).to eq ''
-    end
+    it { expect { normalized_id }.to raise_error Arclight::Exceptions::IDNotFound }
```
drh-stanford: the coverage report says that the `message` isn't being tested here. you can add a second parameter that tests the message.

spec/lib/arclight/normalized_title_spec.rb
```diff
@@ -119,18 +119,6 @@
...
-
-  context 'no information' do
-    let(:title) { nil }
-    let(:date) { nil }
-    let(:id) { nil }
-
-    it 'does not know what to do' do
-      expect(normalized_title).to eq ''
-    end
+    it { expect { normalized_title }.to raise_error Arclight::Exceptions::TitleNotFound }
```
drh-stanford: the coverage report says that the `message` isn't being tested here. you can add a second parameter that tests the message.

mejackreed:
@drh-stanford I've updated the PR addressing your comments.

mejackreed:
I also indexed all of the demo eads using this and didn't raise any errors

jkeck:
Hmm, so if I understand correctly, this will raise an error when there is no title instead of assigning the document id as the title?  Is that the expected behavior @anarchivist?

mejackreed:
Yeah.. that is the new behavior. If the requirement is for each component to have unittitle and unitdate, I think that is approriate

anarchivist:
👍  from me.

PR #451, opened 2017-06-02
fix blacklight title to read arclight

gordonleacock:


lib/generators/arclight/install_generator.rb
```diff
@@ -55,5 +55,9 @@ def add_repository_config
     def add_download_config
       copy_file 'config/downloads.yml' unless File.exist?('config/downloads.yml')
     end
+
+    def modify_blacklight_yml
+      gsub_file 'config/locales/blacklight.yml', "(^.*)\n(.*)\n(.*)", "\1\n\2\n    application_name: 'Arclight'"
```
jkeck: Can we just replace `application_name: 'Blacklight'` with `application_name: 'ArcLight'` instead of using the regular expression?

gordonleacock:
Trying your suggestion @jkeck!

gordonleacock:
def modify_blacklight_yml
      gsub_file 'config/locales/blacklight.yml', 'Blacklight', 'Arclight'
    end

Unfortunately, application_name is still 'Blacklight'.

gordonleacock:
I'm missing an "en" in the file name!

jkeck:
@gordonleacock I would recommend also ensuring that it's `application_name: 'Blacklight' that we're changing.  Just in case more strings end up in that file w/o us knowing about it (that we may not want to change).

gordonleacock:
It's working but I need to add a test.

PR #454, opened 2017-06-09
Update styling of request button and bookmark checkbox

ggeisler:
There are a number of issues we need to work out related to the styling and placement of accessory elements such as the request containers button, the bookmark checkbox, and the view hierarchy toggle. They're all related and somewhat complicated to sort out, so this PR is just a start. It only updates the styling of the request button so that it takes up less room, and places the request button and bookmark checkbox at the right-side of the elements they're in. I'll address the other related issues in future PRs.

Some examples:

### Before - bookmark button inconsistently placed and font larger than it is in other contexts
before-3

-- and -- 
before-1

### After
after-3

----
### Before - request button takes up a lot of space, requires new line for bookmark checkbox
before-2

### After
after-2

### DIFF
--- a/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
@@ -24,6 +24,11 @@
     text-align: right;
   }
 
+  .bookmark-toggle,
+  .al-request-form {
+    font-size: $font-size-sm;
+  }
+
   // Component title + children badge
   .documentHeader {
     border-bottom: $border-width dashed #ccc;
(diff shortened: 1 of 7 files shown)


anarchivist:
Thanks @ggeisler - this definitely makes things look nicer. A couple of points of feedback that may not necessarily blockers to getting this merged. These were both seen in the Alpha Omega Alpha finding aid. Let me know how you'd like to approach this (i.e., new tickets, or addressing this in this PR) - happy to talk face to face if that's easier too.

**Online content label layout; bottom border doesn't extend**

screen shot 2017-06-13 at 3 54 58 pm

**Controls break into two lines inconsistently**

screen shot 2017-06-13 at 3 56 30 pm

anarchivist:
@ggeisler and I chatted about this today and my understanding is that he will address these issues in future PRs.

PR #455, opened 2017-06-12
Split out and relocate the Download select on Collection show page

ggeisler:
Currently, the Download select menu on the collection show page is part of the set of tabs that control your view of the collection:

abdeen_jabara_papers__1956-1994__bulk_1968-1993_-_arclight

----
However, the Download action is very different from those tabs and to prevent confusion should be visually separated. This PR moves the Download button to be right-aligned on the page by taking it out of the current set of nav items:

alpha_omega_alpha_archives__1894-1992_-_arclight
----
At medium and smaller viewports the Download button is bumped down to a separate row to keep the visual separation:

alpha_omega_alpha_archives__1894-1992_-_arclight 2

### DIFF
--- a/app/assets/stylesheets/arclight/modules/show_collection.scss
+++ b/app/assets/stylesheets/arclight/modules/show_collection.scss
@@ -76,3 +76,7 @@
     text-transform: uppercase;
   }
 }
+
+.al-collection-actions-menu {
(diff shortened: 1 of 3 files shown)


anarchivist:
👍  from my perspective.

PR #471, opened 2017-06-22
Fixi misspelled "Michagan"

remocrevo:


mejackreed:
Thanks for catching that!

PR #486, opened 2018-07-12
Updates to try and get builds passing on latest 7.0.0.rc1

jkeck:


spec/features/collection_page_spec.rb
```diff
@@ -263,9 +263,10 @@
       it 'component metadata' do
         within '#contents' do
           within '.document-position-0 ' do
+            puts page.find('.al-document-abstract-or-scope').text
```
mejackreed: Cruft?

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,2 +1 @@
-gem 'blacklight', github: 'projectblacklight/blacklight', branch: 'master'
-gem 'blacklight_range_limit', github: 'projectblacklight/blacklight_range_limit', branch: 'blacklight-7'
+gem 'blacklight_range_limit', '7.0.0.rc2'
```
mejackreed: Do we need this here still since we have an rc?

mejackreed: Ah I guess we do. Seeing that require issue without it 😿

PR #526, opened 2019-07-19
Remove image alt text value

### LINKED ISSUE #525: Use empty value for repository image alt text in "In person" sidebar
For the image we show in the In person sidebar as part of the repository info:

Screen Shot 2019-07-19 at 3 44 49 PM

we currently include an `alt` value with the repository name (i.e., we use `repository.name` to populate the `alt` attribute value). However, since immediately adjacent to that image we display a link with the repository name (populated using the same `repository.name` value), the screen reader user will hear the exact same repository name repeated twice in a row.

So in this context I believe we can consider the image to be decorative and thus not deserving of an `alt` attribute value. It should instead just be an empty value:`<img alt="" class="img-fluid float-left" src="https:` etc.

ggeisler:
Closes #525.


ggeisler:
Deleting because I accidentally included an unnecessary commit in the PR. Will resubmit clean PR.

PR #564, opened 2019-08-14
Add note about Chrome and tests to README

estelendur:
Selenium doesn't work on Linux unless you install Google Chrome, so if you don't install Chrome then the tests will fail for no apparent reason. I figured I should quickly document that.


seanaery:
Thanks @estelendur !

PR #569, opened 2019-08-16
Fix #513 online content ordering

### LINKED ISSUE #513: Sort "Online" items by arrangement order
- [x] Change the sort order of items in the "Online" tab to **arrangement** (order they appear in the "Contents" tab) 

Currently when you select the "Online" tab when on a Collection detail page, the results are displayed in what appears to be alphabetical order:

Screen Shot 2019-07-18 at 2 01 01 PM

Although we might want to give the user some options related to sort order and layout of items in this view (topic for a different ticket), given the archival context the expected default sort order is by arrangement.

estelendur:
Fixes #513.

### DIFF
--- a/app/models/concerns/arclight/search_behavior.rb
+++ b/app/models/concerns/arclight/search_behavior.rb
@@ -26,7 +26,7 @@ module Arclight
     ##
     # For the hierarchy view, set the sort order to preserve the order of components
     def add_hierarchy_sort(solr_params)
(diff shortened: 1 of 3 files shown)


estelendur:
You won't be able to see this in the UI without using data that has a collection with multiple components including online content. Currently none of the fixture data has that property. When I was investigating the issue, I added some strawman fixture data to see it in the UI, but I did not commit any fixture changes.

Theoretically the solr param test that I added should be sufficient to prevent a regression in practice.

PR #576, opened 2019-08-16
Stop online content viewer from reinitializing on second click

### LINKED ISSUE #572: "Online content" reinitializing on second click
## Current Behavior

When on an item that has online content, click the "Online content" loads the content, but clicking overview and then back to "Online content" will reinitialize the content, losing context of where the user was in the digital object viewer.

Example:

* https://arclight-demo.projectblacklight.org/catalog/a0011-xmlaspace_ref6_lx4
* Source data: https://github.com/sul-dlss/arclight-demo/blob/master/data/ead/sul-spec/a0011.xml

bug

## Expected Behavior

Clicking "Online content" the second time should not reinitialize the content. It should be left in the state it was previously.

estelendur:
Resolves #572.

The embedded viewer is fairly opaque to the Javascript, so it seemed most doable to check whether it has content loaded, and if it does, to skip loading it.

I am worried about the brittleness of the fix here. Any thoughts?


mejackreed:
Yeah I see what you mean (about brittleness). What if we dropped a flag in the element so the event only fired once?

estelendur:
Added a flag! Works like a charm.

PR #577, opened 2019-08-19
Show/hide histogram label swaps show and hide with expand/collapse

### LINKED ISSUE #458: Change label of date distribution toggle to reflect current state of toggle
On the search results page, instead of always using the label "Show/hide date distribution" it would be better if the label reflected the current state of the date distribution visualization:

- Visualization is hidden, label says "Show date distribution"
- Visualization is visible, label says "Hide date distribution"

collection__abdeen_jabara_papers__1956-1994__bulk_1968-1993_-_arclight_search_results

estelendur:
Resolves #458.

### DIFF
--- a/app/assets/javascripts/arclight/search_results.js
+++ b/app/assets/javascripts/arclight/search_results.js
@@ -1,0 +1,15 @@
+Blacklight.onLoad(function () {
+    'use strict';
+
+    var histogram_content = $('#al-date-range-histogram-content')
+    var show_span = $('#al-date-range-histogram-show')
+    var hide_span = $('#al-date-range-histogram-hide')
+    histogram_content.on('show.bs.collapse', function() {
+        show_span.attr('hidden', 'hidden')
+        hide_span.removeAttr('hidden')
+    })
+    histogram_content.on('hide.bs.collapse', function() {
+        hide_span.attr('hidden', 'hidden')
+        show_span.removeAttr('hidden')
+    })
+});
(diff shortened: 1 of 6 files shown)


config/locales/arclight.en.yml
```diff
@@ -3,6 +3,8 @@ en:
     masthead_heading: 'Archival Collections at Institution'
     date_range_histogram:
       show_hide: 'Show/hide date distribution'
```
mejackreed: Should we remove this key now?

app/assets/javascripts/arclight/search_results.js
```diff
@@ -0,0 +1,15 @@
+Blacklight.onLoad(function () {
+    'use strict';
```
mejackreed: Can we use 2 space indention here? (sorry its not in the eslint for some reason) Its the convention in all of the other JavaScript.

config/locales/arclight.en.yml
labradford: good catch!

estelendur: Done!

app/assets/javascripts/arclight/search_results.js
estelendur: Yup! Also maybe I should add that to the eslint?

estelendur: It is in the eslint via the included `airbnb-base/legacy`. The eslint was not actually running on my machine, because when I haven't got it installed, the rake task silently fails rather than erroring. I wonder if it's installed and running on Travis.

PR #580, opened 2019-08-19
Remove the link from the title of the current component in context vi…

### LINKED ISSUE #515: Unlink the currently selected component
- [ ] Remove the link applied to the title of the currently selected component in component detail view

On the component detail page we highlight the current component and show some collection context around it.

Screen Shot 2019-07-18 at 2 23 13 PM

It (probably) makes sense for the other component titles shown for context to be linked but the link for the current component simply refreshes the page/context area. This is potentially confusing for the user, who likely would expect to see a different display of some sort when clicking the link of the current component.

seanaery:
…ew. Closes #515.

### DIFF
--- a/spec/features/component_page_spec.rb
+++ b/spec/features/component_page_spec.rb
@@ -131,6 +131,19 @@ RSpec.describe 'Component Page', type: :feature do
       end
     end
     context 'siblings and highlighted self' do
+      it 'does not link to itself' do
+        within '#collection-context' do
+          within '.al-contents' do
+            expect(page).not_to have_css(
+              'article.al-hierarchy-highlight h3 a'
+            )
+            expect(page).to have_css(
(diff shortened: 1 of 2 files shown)


seanaery:
I ran into a Travis CI build issue in the environment using Ruby 2.5.5 ([details](https://travis-ci.org/projectblacklight/arclight/jobs/573975383)) that seems unrelated to the code changes I have made. It might just require rerunning the build, but I'm unable to do that directly.

spec/features/component_page_spec.rb
```diff
@@ -131,6 +131,19 @@
       end
     end
     context 'siblings and highlighted self' do
+      it 'does not link to itself' do
```
mejackreed: Thanks for adding the test

Issue #567, opened 2019-08-16
Add solr_ead fields to traject based indexing


mejackreed:
WIth Traject indexing implemented in #565, we need to add the remaining fields documented in https://github.com/projectblacklight/arclight/issues/557#issuecomment-521801873 to the `ead2_config.rb`. Also checkout `solr_document.rb`

## Collection
 - level - jr
 - repository_(ssm, sim) - jr
 - unitid - jr
## Component
 - parent_ssm - jr
 - parent_unittitles_ssm - jr
 - repository_(ssm, sim) - jr
 - creator_(ssm, ssim, sort) - jr
 - creators_(ssim) - jr
 - collection_creator_ssm - jr
 - parent_access_restrict_ssm
 - parent_access_terms_ssm
 - names_ssim - #595 @seanaery 
 - geogname_(sim, ssm) - #589 
 - places_ssim - jk
 - access_subjects_(ssim, ssm) - @jrgriffiniii (please see https://github.com/projectblacklight/arclight/pull/585)
 - containers_ssim - jk
 - names_coll_ssim - #596 @seanaery 
 - unitid - jr

mejackreed:
Resolved by #666

PR #582, opened 2019-08-19
Adds more traject fields, tests, and fixes

mejackreed:
Part of #567


lib/arclight/traject/ead2_config.rb
```diff
@@ -120,8 +125,26 @@
...
+  end
+
+  to_field 'parent_unittitles_ssm' do |_record, accumulator, context|
+    ## Top level document
+    accumulator.concat context.clipboard[:parent].output_hash['normalized_title_ssm']
+    ## Other components
+    context.output_hash['parent_ssm']&.drop(1)&.each do |id|
+      accumulator.concat Array
+        .wrap(context.clipboard[:parent].output_hash['components'])
+        .find { |c| c['ref_ssi'] == [id] }&.[]('normalized_title_ssm')
```
mejackreed: We need to check `ref_ssi` field as `id` is concated with document id

PR #584, opened 2019-08-20
Add places_ssim to running traject configs PR.

jkeck:


lib/arclight/traject/ead2_config.rb
```diff
@@ -236,7 +239,9 @@
   # to_field 'names_ssim'
   # to_field 'geogname_sim'
   # to_field 'geogname_ssm'
-  # to_field 'places_ssim'
+
+  to_field 'places_ssim', extract_xpath('xmlns:controlaccess/xmlns:geogname')
```
jkeck: We aren't using a fixture in our test currently that has this element at the component level so there's no test for this.  I could refactor the test a bit to allow for that to be more easily done.

PR #585, opened 2019-08-20
Indexing <controlaccess> elements into the access_subjects_ssim and access_subjects_ssm fields

jrgriffiniii:
Resolves #469

Issue #469, opened 2017-06-22
Terms in nested <controlaccess/> elements are not indexing


anarchivist:
Identified in a sample finding aid from @jrees67 (see [here](https://drive.google.com/file/d/0By_BSykwwwI2T1dWSjRERkczV1E/view?usp=sharing)). 

**Sample data**

```xml
<controlaccess>
<head>Index Terms</head>
<p>These terms are indexed in the National Library of Medicine's online catalog LocatorPlus. Researchers wishing to find related materials should search the catalog using these terms.</p>
<controlaccess>
<head>MESH Subjects</head>
  <subject encodinganalog="650$a" source="mesh">Acquired Immunodeficiency Syndrome</subject>
  <subject encodinganalog="650$a" source="mesh">African Americans</subject>
<subject encodinganalog="650$a" source="mesh">Homosexuality</subject>
<subject encodinganalog="650$a" source="mesh">Human Immunodeficiency Virus</subject>
<subject encodinganalog="650$a" source="mesh">Public Health</subject>
</controlaccess>

<controlaccess>
<head>Corporate Names</head>
<corpname encodinganalog="610$a" source="mesh">United States. Presidential Commission on the Human Immunodeficiency Virus  Epidemic</corpname>
</controlaccess>
</controlaccess>
```

**Acceptance criteria**

These should be indexed like any other controlled access terms.

mejackreed:
Preferred to implement w/ traject based indexing strategy.

PR #593, opened 2019-08-20
Index access_subjects at their individual levels (document, component)

mejackreed:
A follow on to #585. I noticed that for each component document, it had all of the other `access_subjects` as part of it.  This still maintains the nesting requested in #469.

However, I realize we are indexing all of the controlaccess fields now. Previously we had specified to constrain this, thus the follow on commit to do such. https://github.com/projectblacklight/arclight/commit/0dd37fd9e6569649fc2ae70b922eb95f5eddfc53

### DIFF
--- a/lib/arclight/traject/ead2_config.rb
+++ b/lib/arclight/traject/ead2_config.rb
@@ -104,6 +104,21 @@ end
 
 to_field 'places_ssim', extract_xpath('//xmlns:archdesc/xmlns:controlaccess/xmlns:geogname')
 
+
+# Indexes the controlled terms for archival description into the access_subject field
+# Please see https://www.loc.gov/ead/tglib/elements/controlaccess.html
+to_field 'access_subjects_ssim', extract_xpath('//xmlns:archdesc/xmlns:controlaccess', to_text: false) do |_record, accumulator|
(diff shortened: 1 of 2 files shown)


lib/arclight/traject/ead2_config.rb
```diff
@@ -80,6 +80,19 @@
 
 to_field 'geogname_sim', extract_xpath('//xmlns:archdesc/xmlns:controlaccess/xmlns:geogname')
 
+# Indexes only specified controlled terms for archival description into the access_subject field
+to_field 'access_subjects_ssim', extract_xpath('//xmlns:archdesc/xmlns:controlaccess', to_text: false) do |_record, accumulator|
+  accumulator.map! do |element|
+    %w[subject function occupation genreform].map do |selector|
```
jrgriffiniii: Apologies for the unnecessary elements in https://github.com/projectblacklight/arclight/pull/585, I wasn't certain if it was desireable to have certain elements (e. g. `corpname`) indexed twice for a document.

mejackreed: Oh its all good. I ran into this while looking at something else. Its confusing to me, but working through this, I think I understand now what we are after.

PR #612, opened 2019-08-22
Add any field defined in the repository.yml for a repository as an …

### LINKED ISSUE #592: Make Arclight::Repository open to extension of new fields
Spun out of #541

Currently, Arclight::Repository has hardcoded accessors. Can we make this extendable so institutions can customize with new fields?

https://github.com/projectblacklight/arclight/blob/f38ac02b7012ade6284b51c47f64c5c88754bdaa/lib/arclight/repository.rb#L10-L25

jkeck:
…accessor

Closes #592

### DIFF
--- a/lib/arclight/repository.rb
+++ b/lib/arclight/repository.rb
@@ -7,31 +7,14 @@ module Arclight
   class Repository
     include ActiveModel::Conversion # for to_partial_path
 
-    FIELDS = %i[name
-                description
-                visit_note
-                building
-                address1
-                address2
-                city
-                state
-                zip
-                country
-                phone
-                contact_info
-                thumbnail_url
-                google_request_url
-                google_request_mappings
-                collection_count].freeze
-
-    attr_accessor :slug, *FIELDS
+    attr_accessor :slug, :collection_count
 
     # @param [String] `slug` the unique identifier for the repository
     # @param [Hash] `data`
     def initialize(slug, data = {})
       @slug = slug
-      FIELDS.each do |field|
-        value = data[field.to_s]
+      data.each do |field, value|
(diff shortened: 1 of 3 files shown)


lib/arclight/repository.rb
```diff
@@ -7,31 +7,14 @@ module Arclight
...
+    attr_accessor :slug, :collection_count
 
     # @param [String] `slug` the unique identifier for the repository
     # @param [Hash] `data`
     def initialize(slug, data = {})
       @slug = slug
-      FIELDS.each do |field|
-        value = data[field.to_s]
+      data.each do |field, value|
+        self.class.attr_accessor field.to_sym
```
jkeck: There is some potential for future race condition issues here that I don't believe we're susceptible to now.  Happy to talk through my concerns further, or just ship it and deal w it if we ever cross that bridge.

mejackreed: Could you just briefly describe the future race condition? So at least we have it shared?

mejackreed: I guess ArcLight could reference a field that hasn't been defined?

E.g

I only provide two fields in my config. But it calls a third not defined?

jkeck: My concern is if we were to do something like initialize additional repositories at runtime (which we don't have the facility to do currently) there could be an issue where we tried to use an accessor that hasn't been defined yet.  Maybe it'll never happen, and because `initialize` is at the class level so it all just works out.  I've just never seen the `self.class.attr_accessor` used like this in the wild so I'm just like, Ruby: 

you cray

mejackreed: 🆒

PR #613, opened 2019-08-22
Run yarn install before build in travis (and error if user does not have eslint installed)

### LINKED ISSUE #581: eslint not running on Travis
Rake silently fails instead of erroring when eslint isn't installed. It looks like the setup tasks aren't running `npm install`, so eslint isn't installed, so it doesn't get run at all, so the javascripts don't get linted.

jkeck:
Closes #581

### DIFF
--- a/tasks/arclight.rake
+++ b/tasks/arclight.rake
@@ -5,6 +5,8 @@ require 'engine_cart/rake_task'
 require 'rspec/core/rake_task'
 require 'arclight'
 
+class DependencyNotInstalled < StandardError; end
+
 desc 'Run test suite'
 task ci: %w[arclight:generate] do
   SolrWrapper.wrap do |solr|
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -0,0 +1,1208 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jkeck: Wasn't sure if this should be gitignored or checked in.  🤷‍♂️

tasks/arclight.rake
```diff
@@ -20,6 +20,7 @@ end
 
 desc 'Run Eslint'
 task :eslint do
+  system 'yarn install'
```
jkeck: It's unfortunate to have to run this on every build (it's relatively fast, but still).  I could imagine doing this on travis in a `before_script` or something like that, but then that would leave the potential for a developer to not have installed the dependencies.  We could put at higher up as part of our install process, which would alleviate this need, but wasn't sure where we would/should slot that in.

yarn.lock
mejackreed: I vote for ignore for a "gem" a la `Gemfile.lock` style. I think we will thrash it around otherwise.

tasks/arclight.rake
estelendur: We could do it on travis in a before_script and then alter the task to fail if eslint is not installed, perhaps?

jkeck: Hmm, interesting idea.  Let me poke at that.

PR #619, opened 2019-08-23
Index collection-level names (corpname, famname, name, persname) via …

### LINKED ISSUE #596: Analysis and addition of field names_coll_ssim  to traject indexing
Field: `names_coll_ssim `

Spun out of #567

## Acceptance criteria
Using work in https://github.com/projectblacklight/arclight/issues/557#issuecomment-521801873, [`solr_document.rb`](https://github.com/projectblacklight/arclight/blob/master/app/models/concerns/arclight/solr_document.rb), and currently indexed fields 

 - [ ] determine if `` is being used
 - [ ] determine if variations should be indexed e.g `_ssm`, `_sim`, `ssim`
 - [ ] determine at what level this should happen. `document`? `component`?, both??
 - [ ] implement in traject config

seanaery:
…Traject. Closes #596.


lib/arclight/traject/ead2_config.rb
```diff
@@ -126,6 +126,14 @@
   accumulator.replace([accumulator.any?])
 end
 
+to_field 'names_coll_ssim', extract_xpath('/xmlns:ead/xmlns:archdesc[@level="collection"]/xmlns:controlaccess', to_text: false) do |_record, accumulator|
```
seanaery: Couple notes on the xpath here:

- uses root-relative expression to get to the `archdesc` rather than `//` as `archdesc` [can only be a child of `ead`](https://www.loc.gov/ead/tglib/elements/archdesc.html). I would imagine we may see performance gains by not using `//` if the element has a predictable path.
- uses `level="collection"` attribute for specifically targeting collection properties. `@level` is required, but [could be something other than `collection`](https://www.loc.gov/ead/tglib/elements/archdesc.html).

seanaery:
Thanks @billdueber -- great idea. I'd also be able to use that constant in the related `names_ssi` indexing (#595), which shares this same list of elements to be considered as names.

seanaery:
Hi @billdueber -- I've refactored the array to a constant [here](https://github.com/projectblacklight/arclight/pull/619/commits/695bab8ba94a91adb066b22a2236e41940c67464) -- is that what you had in mind? Not sure if I've defined it in an optimal place or if there's a more recommended way to do that...

PR #620, opened 2019-08-23
Implements a NokogiriNamespacelessReader class in order to handle processing EAD Documents without namespaces

### LINKED ISSUE #575: Remove namespaces requirement from EAD traject indexing
@anarchivist mentioned in standup that valid EAD2002 may not contain namespaces. Can we account for this in our traject indexing approach

jrgriffiniii:
Resolves #575 using an approach suggested by @cbeer


mejackreed:
Going to hold off on rebasing this until the traject work is finalized. We will rebase and move forward once that is done.

Issue #645, opened 2019-08-27
Tabs and tab contents use 12 columns


jvine:
Once the sidebar is removed, the full width of the show page is available.

- [ ] tabs and `tab-content` should use 12 columns on all devices
- [ ] tabs should use `nav-fill` to fill the width of the page. 2-word labels may wrap within a tab.

Note that component pages will no longer have an Overview tab (#630), and File-level pages will no longer have an Online content tab, so the number of tabs will vary by the current level in the hierarchy.

**Collection desktop**
<img width="971" alt="desktop 4" src="https://user-images.githubusercontent.com/6945691/63732053-231e9f80-c827-11e9-82ca-2c2bd0b6a5cf.png">
**Collection mobile**
<img width="300" alt="mobile 4" src="https://user-images.githubusercontent.com/6945691/63732051-231e9f80-c827-11e9-8de9-f2dc2e2008e5.png">

**Component desktop**
<img width="960" alt="desktop 3" src="https://user-images.githubusercontent.com/6945691/63732094-58c38880-c827-11e9-8959-e08ccdc0db5a.png">
**Component desktop - no online content**
<img width="946" alt="desktop 3 disabled" src="https://user-images.githubusercontent.com/6945691/63732177-c2dc2d80-c827-11e9-9a9c-b1dc4e859335.png">
**Component mobile**
<img width="300" alt="mobile 3" src="https://user-images.githubusercontent.com/6945691/63732093-58c38880-c827-11e9-944e-eac879d4a1e7.png">

**File desktop**
<img width="961" alt="desktop 2" src="https://user-images.githubusercontent.com/6945691/63732109-6da01c00-c827-11e9-9738-b3a2d903f274.png">
**File mobile**
<img width="300" alt="mobile 2" src="https://user-images.githubusercontent.com/6945691/63732110-6da01c00-c827-11e9-8b84-50939ebf11d6.png">

seanaery:
@jvine - The images in this ticket suggest keeping these nav elements as `nav-pills` though most of the related tickets under #632 imply `nav-tabs`. I think these do look better as tabs--just looking to confirm that's the plan. Also, the changes in #638 do create horizontal pills in the Overview tab in mobile view just below these main tabs, which would probably be confusing if it was just two rows of horizontal pills stacked...

jvine:
[...]

Issue #633, opened 2019-08-26
Add the Access tab, move content from the sidebar


jvine:
Part of #632 

- [ ] Add an Access tab as the last item in the tab list. 

### Tab content

Add tab content in the following order. Content is arranged in the standard `<dl>` structure with exceptions noted.

**Collection overview page**
- [ ] RESTRICTIONS
   - from the "Terms & Conditions" accordion section

**Component pages**
- [ ] RESTRICTIONS
   - **component-level restrictions**, if any
   - `<dd>` value is highlighted with a background colour (start with "info" colour; may change)
- [ ] PARENT RESTRICTIONS
   - from the "Terms & Conditions" accordion section
   - restrictions inherited from the parents of this component

**All pages**
- [ ] TERMS OF ACCESS
   - from the "Terms & Conditions" accordion section
- [ ] PREFERRED CITATION
   - from the "How to cite this collection" accordion section
- [ ] LOCATION OF THIS COLLECTION
   - from the "In person" accordion section
   - display the repository image, with the repository address aligned beside it
   - image has a max-width and max-height of 150px
   - repository name is linked to the Arclight repository page
- [ ] BEFORE YOU VISIT
   - from the "In person" accordion section
- [ ] CONTACT 
   - repository phone and email from the "In person" accordion section

### Collection overview
<img width="989" alt="Access collection" src="https://user-images.githubusercontent.com/6945691/63720880-ee015580-c804-11e9-9e76-7861105ec3d4.png">

### Component page

The alert badge on the Access tab will be ticketed separately. 

<img width="993" alt="Access component" src="https://user-images.githubusercontent.com/6945691/63720886-f2c60980-c804-11e9-9dbd-ea82dc9e66c3.png">

labradford:
Closed with PR https://github.com/projectblacklight/arclight/pull/684

PR #690, opened 2019-09-01
Move collection overview nav into overview tab contents. Render colle…

### LINKED ISSUE #638: Move "Navigation overview" into the Overview tab on Collection page
Part C of #632.

The "Navigation overview" section of the sticky sidebar provides jump-to navigation to what can be extremely long content in the Overview tab. The sidebar is being removed; the tab section is being expanded from 9 to 12 columns; this navigation applies _only_ to the Overview tab. So the goal is to move it within the tab, while retaining its jump-to and sticky behaviour.

- [x] split the content of the Overview tab into 2 columns: content (9) and navigation (3).
  - On desktop, the content will be on the left; navigation on the right. 
  - On mobile, the content should wrap _below_ the navigation. 

- [x] move the navigation section into the new column, without its "card" container
  - On desktop, the list elements are stacked, right-aligned
  - On mobile, they are inline, left-aligned, and can wrap if necessary.

- [~] ~navigation section is sticky at the top of the page~ (splitting to another issue 🎟 - cv )

### Desktop
Nav on desktop

### Mobile
Nav on mobile

### Behaviour

(Stickiness of app header and tabs will be addressed separately. This shows the behaviour & stickiness of the nav itself.)

stickynav

seanaery:
…ction show page with 12-col wide tabbed sections & border. Closes #638 & #645.

### DIFF
--- a/app/views/catalog/_collection_overview.html.erb
+++ b/app/views/catalog/_collection_overview.html.erb
@@ -1,7 +1,15 @@
-<h2 class="sr-only">Collection description</h2>
-<% unless blacklight_config.show.metadata_partials.nil? %>
-  <% blacklight_config.show.metadata_partials.each do |metadata| %>
-    <% next unless fields_have_content?(@document, metadata) %>
-    <%= render partial: 'custom_metadata', locals: { document: @document, field_accessor: metadata } %>
-  <% end %>
-<% end %>
+<div class="row">
+  <div class="col-md-3 order-md-10">
+    <%= render partial: 'collection_overview_nav' %>
+  </div>
+
+  <div class="col-md-9 order-md-1">
+    <h2 class="sr-only">Collection description</h2>
+    <% unless blacklight_config.show.metadata_partials.nil? %>
+      <% blacklight_config.show.metadata_partials.each do |metadata| %>
+        <% next unless fields_have_content?(@document, metadata) %>
+        <%= render partial: 'custom_metadata', locals: { document: @document, field_accessor: metadata } %>
+      <% end %>
+    <% end %>
+  </div>
+</div>
--- a/app/views/catalog/_collection_overview.html.erb
+++ b/app/views/catalog/_collection_overview.html.erb
@@ -1,5 +1,5 @@
 <div class="row">
-  <div class="col-md-3 order-md-10">
+  <div class="col-md-3 order-md-2">
     <%= render partial: 'collection_overview_nav' %>
   </div>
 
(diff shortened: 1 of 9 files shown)


seanaery:
TODO:
- coordinate with @labradford to get these changes in sync w/work for #633 that also deals with sidebar removal and tab content.

seanaery:
Screenshots demonstrating revised tab style for main navigation, 12-col page layout, sticky vertical pills on desktop for collection overview, horizontal inline pills in mobile view.

Screen Shot 2019-08-31 at 11 22 39 PM
Screen Shot 2019-08-31 at 11 23 03 PM
Screen Shot 2019-08-31 at 11 23 37 PM

seanaery:
There are two refinements to this work still needed that, after consultation from @jvine , seemed like good candidates to break off into separate tickets. I will create those shortly.

app/views/catalog/_collection_overview.html.erb
```diff
@@ -1,7 +1,15 @@
-<h2 class="sr-only">Collection description</h2>
-<% unless blacklight_config.show.metadata_partials.nil? %>
-  <% blacklight_config.show.metadata_partials.each do |metadata| %>
-    <% next unless fields_have_content?(@document, metadata) %>
-    <%= render partial: 'custom_metadata', locals: { document: @document, field_accessor: metadata } %>
-  <% end %>
-<% end %>
+<div class="row">
+  <div class="col-md-3 order-md-10">
```
camillevilla:
```suggestion
  <div class="col-md-3 order-md-2">
```
This should be ok as an `order-md-2` since this div is a sibling of the collection description. Can we change this class name so it's a little clearer semantically? I was scanning the HTML in-browser and went looking for 9 siblings 😅

app/views/catalog/_show_default.html.erb
```diff
@@ -1,20 +1,20 @@
 <div class='row'>
   <div class='col-md-12'>
-    <ul class='nav nav-pills' role='tablist'>
+    <ul class='nav nav-tabs nav-fill' role='tablist'>
```
camillevilla: Could you add an `aria-label`to describe these tabs? If @jvine sees this before heading out, she can suggest some text here. If not, feel free to use your discretion and we can revise the text once @ggeisler is back.

Here's an example:
https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

jvine: I propose "Context" until we find something better.

app/views/catalog/_collection_overview.html.erb
seanaery: @camillevilla Aha, yes, good eye! This looks like a BS4ism that I was expecting to work like push/pull.

app/views/catalog/_show_default.html.erb
seanaery: @camillevilla Yes, good idea! I'll make a new i18n key for it and share it between the collection & component show pages.

Issue #631, opened 2019-08-26
Add component context/actions box to component show page


ggeisler:
Part of #628.

A new element of the updated component heading area is a box that tells the user which collection the current component is part of and, when available, the name of the container the current component is in. This box also includes global actions that apply to the current component, which includes the bookmark action, and can include a request button and download links. The download links are kind of special and I'll write up those in a separate ticket; they shouldn't affect the basic design of the context/actions box.

## Box placement

- [x] Right-align the box with the right-side of the page content area. 

- [x] Top-align the box with the top of the breadcrumb structure

Something like this:

<img width="1143" alt="Screen Shot 2019-08-26 at 11 55 52 AM" src="https://user-images.githubusercontent.com/101482/63716048-f43e0480-c7f9-11e9-972b-a9ff32cd62c2.png">

## Box contents

There are four potential elements that go in the box:

- Collection ID
- Request button
- Bookmark checkbox
- Current container label

The last one, current container label, is not applicable if the current component page is at the collection level (example A in screenshot below), and it might not always be available at other levels (in which case the box will look like example A even it is at a level below collection). But when we do have the container information, we want to show it in the context box.

<img width="1027" alt="Screen Shot 2019-08-26 at 12 11 09 PM" src="https://user-images.githubusercontent.com/101482/63716540-1be19c80-c7fb-11e9-9773-4c2ca4dfd60a.png">

---

## Box layout

- [x] For the box styling, use (at least for now) the background and border colors currently used for the header of the Page title and  Collection context boxes (see the `al-show-header-section` and `al-document-title-bar` classes):

<img width="844" alt="Screen Shot 2019-08-26 at 12 31 01 PM" src="https://user-images.githubusercontent.com/101482/63717623-88f63180-c7fd-11e9-9899-8a4085d13a73.png">

- [x] Collection ID: use the font styling used in the same examples mentioned above. Left-align in the box.

- [x] If we have a container label to display, add a `|` (vertical bar) followed by a space and then the value of the container label. There shouldn't be a vertical bar if there is no container label to display (as in A in the screenshot).

- [x] Add a request button using `btn btn-secondary btn-sm` styling. Button should be left-aligned with the Collection ID label above it.

- [x] Add the Bookmark checkbox following the request button and some spacing.

[...]

PR #695, opened 2019-09-03
Ticket #631 Add component context/actions box to component show page

gordonleacock:
There are three problems remaining:
1. Sometime the document object passed in to the show_actions_box_default partial has no document.unitid (see screenshot below) so no collection id can be displayed. The ticket assume that is always present. FIXED
2. The request button in the action box goes to an error page. I altered the button styles in view/arclight/requests/_google_form.html.erb because it seemed that the action_box will replace the other request button displays but I didn't make any other changes. FIXED
3. I'm having problems getting the styles for the action box correct. They are close but not right. @ggeisler , please review and suggest changes. GARY SUGGESTS WAITING FOR LATER CLEAN UP

No collection id

Kudos to @jkeck for great help with this PR! :-) And to @labradford for helping me get started!


gordonleacock:
The conflicts noted above are due to a mistaken merge into master of an earlier version of this PR.

mejackreed:
@ggeisler We can alway know which collection it came from, but in this case we use "Collection ID" to be the `unitid`. And unfortunately we can't always count on a `unitid` being available unless we want to enforce that. But I think its optional in the EAD spec.

app/views/catalog/_show_default.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class='row'>
   <div class='col-md-12'>
-    <ul class='nav nav-tabs nav-fill' role='tablist' aria-label='<%= t('arclight.views.show.tablist_nav') %>'>
```
mejackreed: Why have we changed this markup? This seems to be different than the other design tickets?

app/assets/stylesheets/arclight/modules/layout.scss
```diff
@@ -10,6 +10,22 @@
...
+  font-size: $font-size-sm;
+  margin: 0 (-$spacer) $spacer;
+  padding: 3px $result-item-body-padding;
+
+  .btn-secondary, .al-collection-id, .toggle-bookmark {
+    margin-top: $spacer * .5;
+  }
+  
+  .toggle-bookmark {
+      margin: $spacer ($spacer * .5) (-$spacer * .5) 0;
```
mejackreed: Are there some extra spaces in here?

estelendur:
Sometimes the request button does nothing (presumably there is no google form request link) and in that case it should not be present?

app/views/catalog/_show_default.html.erb
gordonleacock: Arg, fixed. :-)

app/assets/stylesheets/arclight/modules/layout.scss
gordonleacock: Stupid sublime3, fixed!

gordonleacock:
Okay, 
1. if there is a document.unitid, I show it
2. if there is a document.unitid and document.containers.present?, I show a vertical bar " | "
3. if there is  document.containers.present?, I show document.containers.join(', ')
and
4. I use requestable = item_requestable?('', { document: document }) to condition the display of
<%= render partial: 'arclight/requests/google_form', locals: { document: document } %>
in the hope that this resolves esty's @estelendur comment.

gordonleacock:
I think I've resolved the conflicts and agree that we should make the mobile styling a separate ticket. @ggeisler said he would come back to the currently styling later but I want to note that now, because because both the collection id and the request button may not always be available, that the current styling I added is probably worse. Sorry Gary! :-|

gordonleacock:
I'm getting errors here that I don't get in my local environment with rspec. @mejackreed?

mejackreed:
Have you tried regenerating the test app?

```
rm -rf .internal_test_app
bundle exec rake engine_cart:generate
```

gordonleacock:
Thanks Jack. Now, I'm just getting one error locally:
Failures:

  1) Arclight::RepositoriesController#show looks up the repository detail page
     Failure/Error: expect(collections.first.unitid).to eq 'MS C 271'

       expected: "MS C 271"
            got: "MS C 544"

       (compared using ==)
     # ./spec/controllers/arclight/repositories_controller_spec.rb:34:in `block (3 levels) in <top (required)>'

Finished in 46.16 seconds (files took 3.29 seconds to load)
443 examples, 1 failure, 9 pending

I think Bill has a fix for this kind of thing using find? Not sure...

gordonleacock:
I changed spec/controllers/arclight/repositories_controller_spec.rb to work, let's see if that breaks travis...

app/views/catalog/_show_collection.html.erb
```diff
@@ -32,23 +32,22 @@
...
-      <div class='tab-pane' id='online-content' role='tabpanel'>
-        <%= render partial: 'collection_online_contents', locals: { document: document }  %>
-      </div>
-      <div class='tab-pane' id='access' role='tabpanel'>
-        <% unless blacklight_config.show.context_access_tab_items.nil? %>
-          <% items = blacklight_config.show.context_access_tab_items.select { |i|  fields_have_content?(@document, i) }  %>
-          <% items.each_with_index do |item, index| %>
-            <%= render partial: 'access_contents', locals: { document: @document, field_accessor: item, card_index: index} %>
-          <% end %>
+     <div class='tab-pane active' id='overview' role='tabpanel'>
```
mejackreed: I mentioned this in slack, but dropping it here for other reviewers. I'm not sure about the indention change here and if its: 
  - needed at all
 - indicates any markup changes

If possible, could we indent this block as it was previously?

PR #718, opened 2019-09-04
Adds parent collection info to grouped results

### LINKED ISSUE #700: List view for grouped search results - B - Collection info box
Split out of #681 

Screen Shot 2019-08-29 at 4 20 50 PM

**B: Collection info box** - The box contains all the collection-related elements to display for each group of results.

- [x] Add shaded box around collection elements. Use the styling currently used for the header portion of the current search result items (CSS class `al-document-title-bar`), except don't include a border.

mejackreed:
Fixes #700 

This should open additional grouped result tickets relying on the parent document.

Screen Shot 2019-09-04 at 3 47 48 PM


app/views/catalog/_group.html.erb
```diff
@@ -0,0 +1,27 @@
+<% # container for all groups in index view -%>
```
mejackreed: A basic port over from Blacklight https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_group.html.erb

This adds some enhancement but likely will be changed as we implement additional designed tickets.

Issue #669, opened 2019-08-29
Update list view of search results


ggeisler:
Part of #667. 

Note that we intend for the arrangement of search result item elements to very closely match the arrangement of elements in the Collection contents/context tab (the tree view of the collection; see #670) and the list of items in the Online items tab (tickets with designs for those forthcoming).

There are a bunch of icons shown in the mockups below. We still need to decide on final icons and define sizes, colors, etc. (#16). I think the recommendations in this ticket can be completed without worrying about icons (with the caveat that some CSS adjustments might be necessary after we add the icons in later), or you could use some placeholder icons.

## Desktop layout of search results

<img width="877" alt="Screen Shot 2019-08-28 at 4 36 48 PM" src="https://user-images.githubusercontent.com/101482/63900031-11b6ce00-c9b4-11e9-8809-6c47cee4f009.png">

---

**A: Result item label** - No change from current, except:

- [x] Prefix the label with an icon conveying whether the item is a collection, a file (lowest-level component), or something in between those two levels.

**B: Result item breadcrumb** - This is similar to the breadcrumb shown in the current search results display. What's changed: 

- [x] Add the repository to the beginning of the breadcrumb
- [x] Move the breadcrumb so it is the last row of the result item entry
- [x] Change the link styling to remove the link color. The breadcrumb elements should still be individually linked as they are now, we're just trying to reduce the amount of blue (future styling updates TBD).

**C: Online content label** - This is new and should match what we do in the collection contents and online contents tabs. Not all items will have this label.

- [x] If the result item has digital content and the item is a file, show a green image icon.
- [x] If the result item is a parent container that either has digital content or contains at least one child with digital content, show the text "Online content" in green (#28A745).

**D: Bookmark checkbox** - We currently show a bookmark checkbox for each result item. All items should continue to have a bookmark checkbox. What's changed:

- [x] Move the bookmark checkbox so it is right-aligned at the edge of the content area, on the same line as the result item label.
- [x] Replace the text "Bookmark" with a ribbon icon.

**E: Container label** - If the result item has a container label:

- [x] Display the label value on the same line as the result item title
- [x] Use a muted text color for the container label: #ADB5BD

See the third result item in the mockup above for the desired order of the container label, online content, and bookmark checkbox elements.

**F: Extent label** - We currently show the value of "extent" as part of the result item metadata. What's changed:

- [x] Move the extent value, if it exists for the result item, to come directly after the result item label, on the same line
[...]

PR #721, opened 2019-09-04
add icon to search results label for item type

### LINKED ISSUE #715: Update list view of search results - A - Result item label
Split out of #669 

**A: Result item label** - No change from current, except:

- [x] Prefix the label with an icon conveying whether the item is a collection, a file (lowest-level component), or something in between those two levels.

labradford:
Closes #715 part of #669 
### Issue
Prefix the label with an icon conveying whether the item is a collection, a file (lowest-level component), or something in between those two levels.

### Solution
- add helper method to determine the icon based on document['level_ssm']. There will be three options: Collection, File and all others
- add icon to _index_header.html.erb partial
- add styling
- add specs for helper and view

### Notes
- Currently uses placeholder icons
- Styling will need to be updated when the icons are provided

### Result
<a href="https://cl.ly/c858e5f4fd48" target="_blank"></a>


app/helpers/arclight_helper.rb
```diff
@@ -113,6 +113,19 @@ def display_size(size)
     size.to_s
   end
 
+  # determine which icon to show in search results header
+  # these icon names will need to be updated when the icons are determined
+  def document_header_icon(document)
+    level = document['level_ssm'] || nil
```
mejackreed: Can we just use the accessor already available on `document`?

```ruby
document.level
```

labradford: updated to use document.level

app/helpers/arclight_helper.rb
```diff
@@ -113,6 +113,19 @@ def display_size(size)
     size.to_s
   end
 
+  # determine which icon to show in search results header
+  # these icon names will need to be updated when the icons are determined
+  def document_header_icon(document)
+    level = document.level || nil
```
mejackreed: I don't think we need to do the `nil` check here anymore as the accessor in `SolrDocument` handles that.

What do you think about switching this to a case statement as opposed to the nested if/else?

labradford: Refactored to case statement.

mejackreed:
@seanaery 

> Use something other than the magnifying glass as the collection icon -- that one is signaling it could be clicked on to power some kind of search functionality.

What would you suggest as a temporary fix? Something that is already available in ArcLight or Blacklight?

labradford:
> Hi @labradford - I think this is a generally a good approach and using the `blacklight_icon` helper is a good way to render SVGs. Even though the icons here are temporary placeholders until the real ones get finalized, what would you think of making these revisions?:
> 
> 1. Use more than two different icons (add at least a third that distinguishes a File from Series/Subseries/etc).
> 2. Use something other than the magnifying glass as the collection icon -- that one is signaling it could be clicked on to power some kind of search functionality.

These were the only icons I could get to render, so this is why I chose them, knowing that we will be adding icons at a later time.

seanaery:
@mejackreed @labradford - I see now there are really only a couple SVGs currently available via Blacklight/ArcLight. I was thinking it'd be easy to grab a free fontawesome SVG like https://fontawesome.com/icons/archive?style=solid and put in the ArcLight assets though looking closer there's a bit of attribution required; that may be unnecessarily complicated for something that will probably be short-lived in the UI.

PR #743, opened 2019-09-09
es6!

### LINKED ISSUE #587: Enable ES6-style classes in Javascript
Per discussion at retro 8/20, it would be great to have some modern JS niceties! There are a couple of ways to approach this. Folks have suggested:
- inserting Babel into the Rails asset pipeline
```
# In environments/production.rb

config.assets.js_compressor = Uglifier.new(harmony: true)
```
- set up [webpacker](https://github.com/rails/webpacker)

estelendur:
We have a working babel pipeline! Also a surprisingly simple one.

Converting the existing javascripts to es6 would be a good opportunistic refactoring task to chip away at separately.

Resolves #587.


lib/arclight/engine.rb
```diff
@@ -14,6 +15,13 @@ module Arclight
   ##
   # This is the defining class for the Arclight Rails Engine
   class Engine < ::Rails::Engine
+    extend ::Sprockets::BumbleD::DSL
```
cbeer: I'm not sure how this all works, but does this inflict a sprockets dependency on downstream consumers, even if they'd prefer to use e.g. webpack? Maybe it's a non-issue until arclight supports webpack, though.

estelendur: I'm not sure how to test that, actually. I believe we get sprockets via a small chain of dependencies.

PR #753, opened 2019-09-10
Traject: index ancestor accessrestrict & userestrict using xpath axes…

### LINKED ISSUE #653: Is there a better way to access parent components for a given element in our traject config?
@estelendur brought this up in https://github.com/projectblacklight/arclight/pull/610/files#r317277920

We should do some analysis and determine if there is a better pattern for this than dropping into the parent document.

### LINKED ISSUE #752: Components with <accessrestrict> display the restrictions twice in a row
A component with its own `<accessrestrict>` is currently displaying those restrictions both as its "Parent Restrictions" as well as its own "Restrictions", which appears unintentional.

The UI specifications in #633 imply changing the indexing for `parent_access_restrict_ssm` so that field only holds restrictions inherited from parent or other ancestor nodes, keeping `accessrestrict_ssm` for storing/showing the restrictions on the current node.

Example component in the demo:
https://arclight-demo.projectblacklight.org/catalog/heschelabrahamaspace_ref1618_lmk

Screen Shot 2019-09-10 at 8 35 56 AM

seanaery:
…. Bugfix for duplicated parent access restrictions on component show page. Closes #653 & fixes #752

### DIFF
--- a/lib/arclight/traject/ead2_config.rb
+++ b/lib/arclight/traject/ead2_config.rb
@@ -299,6 +299,7 @@ compose 'components', ->(record, accumulator, _context) { accumulator.concat rec
         .select { |c| c['ref_ssi'] == [id] }.map { |c| c['normalized_title_ssm'] }.flatten
     end
   end
+
   to_field 'parent_unittitles_teim' do |_record, accumulator, context|
     accumulator.concat context.output_hash['parent_unittitles_ssm']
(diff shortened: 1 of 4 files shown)


seanaery:
I've updated this branch so it now uses the namespaceless xpaths supported via changes merged into master earlier today.

PR #764, opened 2019-09-11
moved container label into header row on search results page

### LINKED ISSUE #711: Update list view of search results - E - Container label
Split out of #669 

**E: Container label** - If the result item has a container label:

- [ ] Display the label value on the same line as the result item title
- [ ] Use a muted text color for the container label: #ADB5BD

See the third result item in the mockup above for the desired order of the container label, online content, and bookmark checkbox elements.

labradford:
closes #711 ref #669 

### Issue
On the search results page:
Container label - If the result item has a container label:
- Display the label value on the same line as the result item title
- Use a muted text color for the container label: #ADB5BD

### Solution
- moved container label into the header row 
- changed text color to #ADB5BD

### Notes
Please ignore styling/placement of bookmark/actions. This will be fixed on a separate ticket.

### Demo
<a href="https://cl.ly/3e032f755622" target="_blank"></a>

### DIFF
--- a/app/assets/stylesheets/arclight/modules/search_results.scss
+++ b/app/assets/stylesheets/arclight/modules/search_results.scss
@@ -28,6 +28,10 @@
       }
     }
 
+    .al-document-container.text-muted {
+      color: #ADB5BD !important;
+    }
+
     .breadcrumb-links,
     .al-document-creator,
     .al-document-extent,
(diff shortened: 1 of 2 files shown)


ggeisler:
Good point about the color contrast issue @seanaery. We have a to-do on our design list to revisit the entire color palette (something I wanted to talk to @mejackreed about) for both consistency and contrast reasons, but this is a good reminder of why we need to do that.

PR #765, opened 2019-09-11
Implements new compact results view

### LINKED ISSUE #671: Update compact view of search results
Part of #667.

It probably makes most sense to work on this ticket after the ticket for the list view of results (#669) has been completed (or work on both at the same time, keeping in mind the differences below), since the compact view is very similar to the list view, just with some subtractions.

## Differences from list view of results

Screen Shot 2019-08-29 at 9 12 31 AM

**A: Omit display of result item metadata** - In list view we show, when available for an item, the scope note or abstract (if a collection)

- [x] In compact view omit the result item metadata.

**B: Truncate breadcrumb** - To keep the view compact, we need to show an abbreviated version of the result item breadcrumb. 

- [x] Display only the first two parts of the item breadcrumb: the repository and the collection. 
- [x] After the collection and the breadcrumb divider icon, show an ellipses as shown in the mockup above. The repository and the collection parts are linked as usual; the ellipses is not linked.

If the result item includes a hit highlighted snippet, it is included with no change from the list view:

Screen Shot 2019-08-29 at 9 25 33 AM

## Mobile
There is no change needed for the compact view of results on mobile; it is just the same mobile adjustments made to _list view_ on mobile, plus the two differences for the compact view described above. So a mobile display of results in the compact view would look something like:

Screen Shot 2019-08-29 at 9 27 41 AM

mejackreed:
Fixes #671 . Split out #766 and #767 once this gets done for list view

Screen Shot 2019-09-11 at 3 04 24 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/search_results.scss
+++ b/app/assets/stylesheets/arclight/modules/search_results.scss
@@ -37,6 +37,12 @@
   }
 }
 
+.documents-compact {
+  h3 {
+    font-size: $h5-font-size;
+  }
+}
+
 .al-grouped-results {
   .al-document-abstract-or-scope {
     max-width: 45em;
--- a/app/helpers/arclight_helper.rb
+++ b/app/helpers/arclight_helper.rb
@@ -18,6 +18,30 @@ module ArclightHelper
   end
 
   ##
+  # For a non-grouped compact view, display the breadcrumbs with the following
+  # algorithm:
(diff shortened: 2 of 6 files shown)


ggeisler:
Not sure if this is known and will be dealt with separately, but if not, the hit highlighting line should come before the breadcrumb. See "B" in #671 for a visual example.

mejackreed:
Thanks for that @ggeisler .. I've updated it.

PR #775, opened 2019-09-12
[WIP] Add custom pagination information for grouped responses fixes #699

### LINKED ISSUE #699: List view for grouped search results - A - Update the result count message
Split out of #681 

Screen Shot 2019-08-29 at 4 20 50 PM

 **A: Update the result count message.** Since the user is interested in seeing results via collections, the message should reflect the number of **collections** returned by the query, instead of documents (the number of docs per collection is covered in F). Showing 10 collections per page, each with several docs, does seem like it will create a large page (vertically and page-load wise), but we have the per-page dropdown to consider, and we're not sure how easily we can operate with two per page approaches in the same app.

- [ ] Display a result count message in the format "1 - x of y collections" where x uses the per-page parameter (based on the number of groups/collections) and y is the total number of groups/collections returned by the grouped query.

mejackreed:
Fixes #699 
The `#page_entries_info` approach is a little gnarly. I'm open to alternative ways to tackle this if others have suggestions?

Screen Shot 2019-09-12 at 3 48 29 PM
Screen Shot 2019-09-12 at 3 48 00 PM


app/helpers/blacklight_helper.rb
```diff
@@ -0,0 +1,38 @@
+module BlacklightHelper
```
cbeer: I'm not sure this is how we want to expose this behavior out of the arclight engine (for one, if the app has its own `BlacklightHelper`, we won't get this behavior). You might be better off putting the override  in its own helper and having the install generator inject it into the application.

Or, this seems like something that could get ported upstream, perhaps?

.rubocop_todo.yml
```diff
@@ -192,6 +192,14 @@ Style/ExpandPathArguments:
     - 'lib/generators/arclight/install_generator.rb'
     - 'lib/generators/arclight/update_generator.rb'
 
+# Offense count: 1
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle.
+# SupportedStyles: always, never
+Style/FrozenStringLiteralComment:
```
cbeer: Depending on the resolution of my other comment, should we just go ahead and add the magic comment (or disable it in the regular rubocop file)?

mejackreed: 👍 gotcha.. yeah updated and force pushed.

app/helpers/blacklight_helper.rb
mejackreed: It is unclear to me what the best way to handle this is.

The logic _could_ be ported upstream to Blacklight. But my understanding is that Blacklight's `page_entries_info` seems to be handing some cases that may not exist anymore. Like this `model_name` stuff? https://github.com/projectblacklight/blacklight/blob/master/app/helpers/blacklight/catalog_helper_behavior.rb#L44-L47

Another hesitation to push this upstream, was where and how we would be able to pass `entry_name` for when the context is a grouped result. ArcLight turns on grouping as a "mode" and Blacklight doesn't really have that understanding as far as I can tell.

Open to suggestions on where we might do this.

mejackreed:
Per @cbeer suggestion, I'll be looking to make some changes upstream in Blacklight.

mejackreed:
Closing in favor of https://github.com/projectblacklight/blacklight/pull/2140

PR #777, opened 2019-09-13
Allow for download URLs to be configured in a programatic way

### LINKED ISSUE #683: Define a programmatic approach to constructing download links
`downloads.yml` already provides a useful mechanism for a repository to define individual downloads for a given collection-level description. Extend this to enable the following:

 - [x] a programmatic construction of a url 

The intent is to allow ArcLight implementers to customize this locally, add custom logic, call an external API, etc., as they need to.

jkeck:
Closes #683 

The download links (which are not currently rendered anywhere) can now be rendered by executing this line:

```ruby
render Arclight::CollectionDownloads.new(document)
```

If/when this is merged I'll add a wiki page about configuring download links with more info.

### DIFF
--- a/app/models/arclight/collection_downloads.rb
+++ b/app/models/arclight/collection_downloads.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+module Arclight
+  ##
+  # Model the Download links that can be configured (via YAML) for a collection
+  class CollectionDownloads
+    attr_reader :document
+
+    def initialize(document)
+      @document = document
+    end
+
+    def id
+      document.unitid
+    end
+
+    def files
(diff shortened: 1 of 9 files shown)


app/models/arclight/collection_downloads.rb
```diff
@@ -0,0 +1,112 @@
...
+      end.compact
+    end
+
+    def to_partial_path
+      'catalog/collection_downloads'
+    end
+
+    private
+
+    def config
```
cbeer: Are there future plans to make this config document-specific, or should it get extracted to the class level (or some other place?) so we're not continually re-parsing it?

jkeck: Decided to pull this up the class level.

PR #794, opened 2019-09-16
709 online content

### LINKED ISSUE #709: Update list view of search results - C - Online content label
Split out of #669 

**C: Online content label** - This is new and should match what we do in the collection contents and online contents tabs. Not all items will have this label.

- [x] If the result item has digital content and the item is a file, show a green image icon.
- [x] If the result item is a parent container that either has digital content or contains at least one child with digital content, show the text "Online content" in green (#28A745).

labradford:
closes #709 ref: #669 

### Issue:
If the result item has digital content and the item is a file, show a green image icon.
If the result item is a parent container that either has digital content or contains at least one child with digital content, show the text "Online content" in green (#28A745).

### Discussion:
- Q - When displaying the "online content" label, are we using the same logic as before with the exception of showing an icon when the item is of the type file?
- A - Yep, logic is unchanged, just replacing the "online content" label with the icon.

### Result/Demo:
<a href="https://cl.ly/fa05dab4712a" target="_blank"></a>


ggeisler:
@labradford My answer to you above is quoted correctly but I don't think I interpreted your question the way you intended it. I meant, yes we are using the same logic to display the indicator, but we are replacing the text label indicator with the icon indicator. So there should never be a text label.

I see now that you were asking the question in a slightly different way, which resulted in you only using the icon in some places. Sorry for the confusion.

You might want to wait until later tomorrow to update this PR, since @jvine had some questions about the logic (which I suggested we deal with separately from just replacing the text label with the icon) that she might want to explain if you have to go back and adjust this work anyway.

labradford:
Closing in favor of PR which will update all online content indicators to use the icon.

PR #798, opened 2019-09-17
Remove page params for grouping/ungrouping

mejackreed:
I noticed that grouping/ungrouping can get into a strange state if page params are persisted. This PR removes them.

### DIFF
--- a/app/helpers/arclight_helper.rb
+++ b/app/helpers/arclight_helper.rb
@@ -110,11 +110,11 @@ module ArclightHelper
   end
 
   def search_with_group
-    search_state.params_for_search.merge('group' => 'true')
+    search_state.params_for_search.merge('group' => 'true').reject { |k| k == 'page' }
   end
 
   def search_without_group
-    search_state.params_for_search.reject { |k| k == 'group' }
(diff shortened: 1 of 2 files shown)


app/helpers/arclight_helper.rb
```diff
@@ -110,11 +110,11 @@ def grouped?
   end
 
   def search_with_group
-    search_state.params_for_search.merge('group' => 'true')
+    search_state.params_for_search.merge('group' => 'true').reject { |k| k == 'page' }
```
cbeer: Can we use [`#except`](https://apidock.com/rails/Hash/except) here?
```suggestion
    search_state.params_for_search.merge('group' => 'true').except('page')
```

app/helpers/arclight_helper.rb
```diff
@@ -110,11 +110,11 @@ def grouped?
...
   end
 
   def search_without_group
-    search_state.params_for_search.reject { |k| k == 'group' }
+    search_state.params_for_search.reject { |k| %w[group page].include? k }
```
cbeer:
```suggestion
    search_state.params_for_search.except('group,' 'page')
```

PR #800, opened 2019-09-17
Update breadcrumbs for collection and component show pages #796

### LINKED ISSUE #796: Update breadcrumbs for collection and component show pages
Part of the goal of updating the page header area of collection and component show pages (mostly represented by #628) is to show the breadcrumb hierarchy as one that represents the _repository > collection > component_ context of the current item, by using the breadcrumbs to show that hierarchy. Currently we do some of that but also include the site navigation hierarchy as well:

Screen Shot 2019-09-17 at 9 08 39 AM

Note in the screenshot above that the breadcrumb hierarchy starts with "Home" and "Collections". When the user is on a collection or components how page, we want to substitute the parent repository name for those two elements.

For collection and component show pages:

- [ ] Remove the  "Home" and "Collections" elements from the breadcrumb hierarchy
- [ ] Start the hierarchy with the owning repository of the current item

So the example above should look like this:

Screen Shot 2019-09-17 at 9 16 36 AM

(We also want to replace the `»`s with icons, but I'm working on a separate ticket for that.)

Site pages that are not collection and component show pages (search results, the repositories landing page) should keep their current approach to breadcrumbs (single-line "Home » Search results", etc.).

gordonleacock:
Fixes #796 

Updates the partial app/views/shared/_show_breadcrumbs.html.erb which is used for the breadcrumbs for the collection and component pages. Note that if we are unable to get a document.repository_config from the document that the text "Respositories" is substituted for the repository link. May not happen and may not be the best solution?

Updated specs to match the show_breadcrumbs changes:

- spec/features/collection_page_spec.rb
- spec/features/component_page_spec.rb
- spec/features/item_breadcrumbs_spec.rb

One odd issue with the specs was that I visually counted one more breadcrumb link than the spec insisted where present. For example, in the spec/features/component_page_spec.rb L243, "        expect(page).to have_css 'a', count: 3" works but I saw four breadcrumbs links on the page. So, I'm not sure what is up with that!


ggeisler:
@gordonleacock 
> Note that if we are unable to get a document.repository_config from the document that the text "Respositories" is substituted for the repository link. May not happen and may not be the best solution?

I hope we can always get the repository name, but seems like a good idea to have a fallback in case we can't. I suggest going with "Repository" (singular) instead, though. Product owners might be better suited than us to provide the ideal fallback term, if we think this is really something that will happen.

app/views/shared/_show_breadcrumbs.html.erb
```diff
@@ -1,19 +1,20 @@
...
-    <li class="breadcrumb-item breadcrumb-item-1">
-      <%= link_to t('arclight.routes.home'), root_path %>
-    </li>
-    <li class="breadcrumb-item breadcrumb-item-2 al-bc-item">
-      <%= link_to t('arclight.routes.collections'), arclight_engine.collections_path %>
+    <li class="breadcrumb-item al-bc-item breadcrumb-item-1">
+      <% if document.repository_config.present? %>
+        <%= link_to(document.repository_config.name, arclight_engine.repository_path(document.repository_config.slug)) %>
+      <% else %>
+          <span>Repositories</span>
```
jkeck: This string should be internationalized (and singular per @ggeisler's comment).

gordonleacock: Pushed up these changes and waiting on CI.

PR #807, opened 2019-09-18
Bugfix: add IDs to components missing them; add an id-less fixture; improve test coverage for minting IDs. 

### LINKED ISSUE #785: Component IDs are not getting set correctly when missing
Whether you browse the repositories page or search for known items (Douglas MacArthur, Tinder, 
 Hemingway) I'm not finding anything from cmu-clarke or umich-wlcl. 

The files are there in arclight-demo/data/ead.  These are non-ASpace, EAD 2002 DTD finding aids so may be outliers with new indexing issues.

seanaery:
Fixes #785 . Changes in this PR include:
- Adds a test fixture sample EAD that lacks component IDs. For this I used a truncated/revised version (< 200 lines of EAD) of the CMU Hemingway finding aid mentioned in #785 as that has a few characteristics not yet accounted for in our fixtures (notably: missing IDs, and `archdesc[@level="recordgrp"]`). I also had to convert this finding aid from using EAD2002 DTD to using EAD2002 W3C Schema (XSD), else it [would fail validation in our CI pipeline](https://github.com/projectblacklight/arclight/commit/33dbee3fd39ea7b1a2822a05b93dc29610c130db#diff-354f30a63fb0907d4ad57269548329e3).
- Addresses the bug preventing the component IDs from being written to the index once minted
- Adds tests to verify minted IDs get written correctly
- Revises existing test example for ID hashing so it's valid EAD (note addition of `<dsc>` between the `<archdesc>` & `<c>` elements).
- Improves resiliency for indexing levels & parent IDs by adding a few safe navigation operators.

### DIFF
--- a/lib/arclight/traject/ead2_config.rb
+++ b/lib/arclight/traject/ead2_config.rb
@@ -230,10 +230,11 @@ compose 'components', ->(record, accumulator, _context) { accumulator.concat rec
                      parent_id = context.clipboard[:parent].output_hash['id'].first
                      logger.warn('MISSING ID WARNING') do
                        [
-                         "A component in #{parent_id} did not have and ID so one was minted using the #{strategy} strategy.",
+                         "A component in #{parent_id} did not have an ID so one was minted using the #{strategy} strategy.",
                          "The ID of this document will be #{parent_id}#{hexdigest}."
                        ].join(' ')
                      end
+                     hexdigest
                    else
                      record.attribute('id')&.value&.strip&.gsub('.', '-')
                    end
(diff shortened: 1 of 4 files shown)


lib/arclight/traject/ead2_config.rb
```diff
@@ -230,10 +230,11 @@
                      parent_id = context.clipboard[:parent].output_hash['id'].first
                      logger.warn('MISSING ID WARNING') do
                        [
-                         "A component in #{parent_id} did not have and ID so one was minted using the #{strategy} strategy.",
+                         "A component in #{parent_id} did not have an ID so one was minted using the #{strategy} strategy.",
                          "The ID of this document will be #{parent_id}#{hexdigest}."
                        ].join(' ')
                      end
+                     hexdigest
```
jrgriffiniii: I'm not aware of the discussions which have taken place regarding the minting of IDs using the digest, hence I am hesitant to merge this myself.

mejackreed: 👍 Gotcha.. good catch @seanaery  looks like we just needed to return it

Issue #805, opened 2019-09-18
Add component context/actions box to collection show page


camillevilla:
Descends from #631. Most of this was implemented in #695 for component show pages, so this ticket should hopefully be able to reuse the partial implemented there or refactor it to make it more generic (maybe using the `collection_unitid` implemented in #750?) 

<img width="1143" alt="Screen Shot 2019-08-26 at 11 55 52 AM" src="https://user-images.githubusercontent.com/101482/63716048-f43e0480-c7f9-11e9-972b-a9ff32cd62c2.png">

See case A:
<img width="1027" alt="Screen Shot 2019-08-26 at 12 11 09 PM" src="https://user-images.githubusercontent.com/101482/63716540-1be19c80-c7fb-11e9-9773-4c2ca4dfd60a.png">

mejackreed:
This seems to be resolved now in `master`

Issue #640, opened 2019-08-26
Add download links to component context/actions box when appropriate


ggeisler:
Part of #628.

When a collection's finding aid is available as a PDF and/or EAD xml file, we should display links in the component context/actions box at all component levels for that collection. The link(s) go at the bottom of the box, below the Request and Bookmark elements:

<img width="896" alt="Screen Shot 2019-08-26 at 4 18 21 PM" src="https://user-images.githubusercontent.com/101482/63729709-2b71dd00-c81d-11e9-9375-278c11b5c367.png">

The three examples above show how things look when there is A) only a PDF available, B) only an EAD file available, and C) when both files are available. 

When a collection doesn't have either a PDF and/or EAD xml file available to download, we omit those links entirely from the context/actions box.

## Previous related work

In the previous workcycle work was done that should make displaying the download links relatively easy(?) but as far as I know the requirements are not well documented and I'm not sure we actually have a working example of download links in our current demo instance. So here is some background in case it helps.

We added a feature that enables us to present links for downloading a collection PDF or EAD if one or both of those files have been named and stored appropriately. More specifically, based on comments by the developer who worked on ticket #371 I think the following rules apply:

- A finding aid PDF has to be named `unitid.pdf`, where `unitid` is the collection ID, I assume
- A finding aid EAD has to be named `unitid.xml`
- If either of those files exist and are intended for users to download, they need to be placed in the public/documents directory

Also, based on comments in PR #425, which closed #371, I think these other rules might apply:

- If either a PDF or EAD is intended to be downloadable, it needs to also be documented in a `/config/downloads.yml` file like this:

```
sample_unitid:
  pdf:
    href: 'http://example.com/sample.pdf'
    size: '1.23MB'
  ead:
    href: 'http://example.com/sample.xml'
    size: 123456
```
- Because we have changed the way we want the download links to be available in the UI, we'll have to also update the locale strings in `config/locales/arclight.en.yml` to reflect the new strings used in the mockups above:

```
        download:
          default: 'Download'
          pdf: 'Collection PDF (%{size})'
          ead: 'Collection EAD (%{size})'
```

Note that I'm not sure we ever created real examples in the demo instance of ArcLight to see how this all works in practice.

### Icons

Icon | Source | SVG code
------------ | ----------| -------------
[...]

PR #750, opened 2019-09-10
Generate an Aeon request link using a web-hosted EAD file

camillevilla:
Closes #643
Paired w/ @mejackreed on this.

This PR:
- makes the parent's `unitid` available in the document
- adds a method for rendering request button(s) based off a repository's configured request types
- adds a class for handling Aeon requests w/ a web-hosted EAD 
Sample fixture: 
http://localhost:3000/catalog/umich-bhl-2016071

PR #808, opened 2019-09-18
Introduces download links for Collections and Containers in the action box

jrgriffiniii:
Advances #640 by addressing the following:
- Introduces EAD and PDF download icons
- Restructuring the CollectionDownloads as the DocumentDownloads (as this must be now generalized for both Collections and Containers)
- Ensuring that document downloads are rendered in the action box for Collections
- Updating the locale for the download link text content

Please see the following screenshots for containers:

image
...and collections:
image


app/views/catalog/_show_actions_box_collection.html.erb
```diff
@@ -1 +1,40 @@
...
+    <% end %>
+  </div>
+
+    <% document_downloads = Arclight::DocumentDownloads.new(document) %>
+    <% if document_downloads.files.present? %>
+      <div class="mr-auto al-show-actions-box-downloads">
+
+        <% document_downloads.files.each do |file| %>
+          <div class="al-show-actions-box-downloads-file">
+            <% if file.pdf? %>
```
mejackreed: Instead of the conditional checks here, could we do something like this?

```ruby
<%= blacklight_icon(file.type, classes: "al-show-actions-box-downloads-file") %>
```

app/views/catalog/_show_actions_box_default.html.erb
```diff
@@ -25,4 +25,27 @@
     </div>
   <% end %>
 
+    <% document_downloads = Arclight::DocumentDownloads.new(document) %>
+    <% if document_downloads.files.present? %>
+      <div class="mr-auto al-show-actions-box-downloads">
+
+        <% document_downloads.files.each do |file| %>
+          <div class="al-show-actions-box-downloads-file">
+            <% if file.pdf? %>
```
mejackreed: Similar to comment in `_show_actions_box_collection.html.erb`

lib/arclight/engine.rb
```diff
@@ -46,5 +46,9 @@ class Engine < ::Rails::Engine
     initializer 'arclight.views' do
       Blacklight::Configuration.default_values[:view].hierarchy
     end
+
+    initializer 'arclight.assets.precompile' do |app|
+      app.config.assets.precompile += %w[blacklight/ead.svg blacklight/pdf.svg] if defined? Sprockets
```
mejackreed: When using the `blacklight_icon` method, do we still need to do this? We haven't needed to for other added icons. Since `Blacklight::Icon` reads the file directly from source, I don't think it needs to touch the asset pipeline.

app/assets/stylesheets/arclight/modules/layout.scss
```diff
@@ -16,6 +16,21 @@
   font-size: $font-size-sm;
   height: 100%;
   padding: $spacer;
+
+  &-downloads {
+    padding: 0.32rem;
```
mejackreed: The mocks seem to indicate download links aligned with the request button. Can we update some of the padding here to reflect that. And if its possible, can we use units based off `$spacing` or other defined variables where possible?

Screen Shot 2019-09-18 at 7 43 02 AM
Screen Shot 2019-09-18 at 7 42 43 AM

app/models/arclight/document_downloads.rb
```diff
@@ -3,28 +3,35 @@
 module Arclight
   ##
   # Model the Download links that can be configured (via YAML) for a collection
-  class CollectionDownloads
+  # or container
+  class DocumentDownloads
```
mejackreed: @camillevilla Just as a heads up this has implications on #750

app/views/catalog/_document_downloads.html.erb
```diff
@@ -0,0 +1,16 @@
+<% if document_downloads.files.present? %>
```
mejackreed: Sorry to be daft here.. where is this being rendered? I couldn't figure out how to view this.

app/views/catalog/_show_actions_box_collection.html.erb
```diff
@@ -1 +1,40 @@
-<% # intentionally left blank to not render the actions box for collections %>
+<div class='al-show-actions-box col-md-3'>
```
mejackreed: Is it possible that we could get away without this partial and just use the default one? Requestability logic should constrain the request button from being shown here I think.

lib/arclight/engine.rb
jrgriffiniii: I am finding that errors are raised for me locally when I attempt to remove this, but I will test this on Travis CI.

app/models/arclight/document_downloads.rb
mejackreed: @jkeck Do you have any thoughts on the name of this?

jkeck: I'm 👍 to `Document`.

app/views/catalog/_collection_downloads.html.erb
```diff
@@ -1,16 +0,0 @@
-<% if collection_downloads.files.present? %>
```
jkeck: I'm a bit surprised to see this file be deleted.  The reason that this file is referenced as the `to_partial_path` of the (now) `DocumentDownloads` class is so that we can do this in a view (instead of putting this code explicitly in `_show_actions_box`):

```ruby
render Arclight::DocumentDownloads.new(document) 
```
(or perhaps even better having an accessor on the `SolrDocument` class that returns the `CollectionDownloads` instance for that document and then you can just `render document.download_links` or something along those lines)

This would have the benefit of making this code in the view a bit easier to read and refactor/move later if necessary.

app/views/catalog/_document_downloads.html.erb
jrgriffiniii: You're not daft, I didn't properly test for this.  Following https://github.com/projectblacklight/arclight/pull/808#discussion_r325833504, I realized what could be done to extend the View test suite.

jrgriffiniii:
I have introduced adjustments to the markup and SCSS rules which bring the action box into greater alignment with the wireframes:

image
image
image

jrgriffiniii:
This should also resolve #805

app/views/catalog/_show_actions_box_default.html.erb
```diff
@@ -15,14 +15,33 @@
...
-  <% if requestable %>
-    <div class='mr-auto al-action-box-request'>
-        <%= render partial: 'arclight/requests/google_form', locals: { document: document } %>
+    <div class='mr-auto al-show-actions-box-bookmarks'>
+      <%= render partial: 'catalog/bookmark_control', locals: { document: document } %>
     </div>
-  <% end %>
+  </div>
+
+    <% if document.downloads.files.present? %>
```
jkeck: Thanks for taking care of this James.  One thing that I think I may not have been clear on, is that I think the content that is here in this partial to render the download links can actually be the content of the `document_downloads` partial instead.  Then in this partial, we can simply call `render document.downloads`.  The split button stuff that is in that partial currently isn't necessary w/ the new designs so that can be dropped and that partial can just contain the link rendering markup below.

jrgriffiniii: I'm extremely sorry for that error, this has been adjusted.

app/views/catalog/_show_actions_box_default.html.erb
```diff
@@ -15,14 +15,16 @@
...
-  <% if requestable %>
-    <div class='mr-auto al-action-box-request'>
-        <%= render partial: 'arclight/requests/google_form', locals: { document: document } %>
+    <div class='mr-auto al-show-actions-box-bookmarks'>
+      <%= render partial: 'catalog/bookmark_control', locals: { document: document } %>
     </div>
-  <% end %>
-
+  </div>
+  <%= render document.downloads %>
```
jkeck: 👌

app/assets/stylesheets/arclight/modules/layout.scss
```diff
@@ -16,6 +21,32 @@
...
+    padding-bottom: 0.08rem;
+  }
+
+  &-downloads {
+    padding: 0.32rem 0;
+
+    &-file {
+      padding: 0.24rem 0;
+      margin-left: -0.06rem;
+      margin-right: -0.06rem;
```
jkeck: I'm a bit curious at some of this fine-grained padding.  Is this just doing some pixel-pushing to make things look right, or has this been specified somewhere?  I wonder if there are some things built into Bootstrap that we can/should be looking towards instead (but admittedly can't offer what exactly that would be).

jrgriffiniii: I was definitely struggling with this, I can try and restructure these to use the utility classes from https://getbootstrap.com/docs/4.3/utilities/spacing/#notation

jrgriffiniii:
`Faraday::ConnectionFailed: execution expired` keeps getting issued by Travis CI builds.

PR #823, opened 2019-09-18
Add badge to show number of children in document hierarchy

### LINKED ISSUE #814: Collection context: badge indicating the number of direct children
Split out of #668 / #670

## Components that have children

- [ ]  badge indicating the number of direct children of the component, displayed inline with the title
   - badge is not part of the title link

**Large**
component w children wide
**Medium**
component w children medium
**Small**
component w children small

## File-level components that do not have children
- [ ] No badge should be displayed.

**Large**
file component large
**Medium**
file component medium
**Small**
file component small

estelendur:
Closes #814.

The badge text scales in size with the text it's next to. It may want more specific styling to better match the mockups.

image

image

### DIFF
--- a/app/views/catalog/_index_header_hierarchy_default.html.erb
+++ b/app/views/catalog/_index_header_hierarchy_default.html.erb
@@ -9,12 +9,12 @@
     <% counter = document_counter_with_offset(document_counter) %>
     <%= link_to_document document, document_show_link_field(document), counter: counter %>
     <% if document.children? %>
-      <span class="badge badge-pill badge-secondary"><%= document.number_of_children %></span>
+      <span class="badge badge-pill badge-secondary al-number-of-children-badge"><%= document.number_of_children %></span>
     <% end %>
     <%= render_document_partial(document, 'arclight_online_content_indicator') %>
     <% if document.children? %>
       <div class='al-hierarchy-children-status'>
-        <span class='badge badge-default al-number-of-children-badge'>
+        <span class='badge badge-default'>
           <% unless hierarchy_component_context? %>
             <%= link_to(
               t('arclight.hierarchy.view_all'),
--- a/app/views/catalog/_index_header_hierarchy_default.html.erb
+++ b/app/views/catalog/_index_header_hierarchy_default.html.erb
@@ -9,7 +9,7 @@
     <% counter = document_counter_with_offset(document_counter) %>
     <%= link_to_document document, document_show_link_field(document), counter: counter %>
     <% if document.children? %>
-      <span class="badge badge-pill badge-secondary al-number-of-children-badge"><%= document.number_of_children %></span>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_index_header_hierarchy_default.html.erb
```diff
@@ -8,11 +8,13 @@
     <% end %>
     <% counter = document_counter_with_offset(document_counter) %>
     <%= link_to_document document, document_show_link_field(document), counter: counter %>
+    <% if document.children? %>
+      <span class="badge badge-pill badge-secondary al-number-of-children-badge"><%= document.number_of_children %></span>
```
cbeer: Maybe this is a question for @jvine or @ggeisler, but for accessibility reasons, do we need to provide additional context for this number, akin to what we were doing before (1 child, 15 children, etc)?

estelendur: Probably! I'm not sure what the best way to do that is, exactly.

anarchivist: My understanding from @ggeisler and @jvine was that not including "children" specifically was part of the design decision. We do perhaps need context, but I'll let them mull over how to specify it.

ggeisler: I think @cbeer has a good point. We do intentionally want the visible badge to simply show the number of children, but for a screenreader user, hearing the number alone might not be enough context.

I believe we could add `<span class="sr-only">children</span>` after the number but within the badge span so nothing visibly changes in the UI, but screenreaders would say "3 children":

```
<span class="badge badge-pill badge-secondary al-number-of-children-badge"><%= document.number_of_children %><span class="sr-only">children</span></span>
```
(though I think we also want to use the pluralize method for the "children" text in case there is only one)

We probably should get confirmation from @jvine that this makes sense before we go with it, since she is more of an a11y expert than I am.

jvine: Was trying to resist this, but Chris is right, we should provide the context. We explicitly removed the word "children" from the UI because of usability feedback. Embedding the same word in "invisible" content will make it less likely that implementers will know it's in there creating confusion. 

Suggestion 1: be explicit about "3 folders" for sub-containers and "3 items" for files. a) do we know? and b) can a parent contain both?

Suggestion 2: "components". Feels like it might be jargon-y because we use it internally, but my independent "what word should I use" consultant came up with it on their own, so maybe?

ggeisler: Good point about the term "children". I was just focused on technical implementation and forgot the term "children" was problematic.

cbeer: WFM; it seems like there's enough open questions around suggestion 1 that it could spin off to a new ticket if it's something we want to pursue.

PR #825, opened 2019-09-18
Clean up existing solr configuration

cbeer:
This removes some of the boilerplate that is unused and/or legacy solr configurations that now have better alternatives.


app/models/concerns/arclight/search_behavior.rb
```diff
@@ -47,17 +47,8 @@ def add_highlighting(solr_params)
     ##
     # Adds grouping parameters for Solr if enabled
     def add_grouping(solr_params)
-      if blacklight_params[:group] == 'true'
-        solr_params[:group] = true
-        solr_params['group.field'] = 'collection_ssi'
-        solr_params['group.ngroups'] = true
```
mejackreed: This seems great adding in a config! 👍

PR #844, opened 2019-09-19
Use _nest_path_ instead of sort_ii for figuring out the original order..

cbeer:
… of components within a document.

`sort_ii` indexing was dropped in the switch to traject, and seems redundant with the `_nest_path_` information (which contains e.g. `/component#5` for the 6th (5th? i've assumed it's 0-indexed, but it doesn't really matter..) child of the collection)


cbeer:
(For the record, this doesn't work.. `_nest_path_` doesn't sort the indexes numerically (ugh.).

PR #856, opened 2019-09-20
Build nested lists of parents that retrieve tree structu…

### LINKED ISSUE #818: SPIKE: What will it take to shoehorn our current collection context navigation approach to new updated design
The updated collection context design #670, #673 call for significant changes to how we are displaying the collection context. This ticket's acceptance criteria are as follows:

 - Can our current approach of using `collection_navigation.js` and `component_ancestors.js` be used to accomodate the new designs?
 - If so what are the changes that need to happen in those JS files or elsewhere to make this possible
 - If not, what are the barriers?
 - Does any indexed data need to be changed or updated?

mejackreed:
…re of ead

Screen Shot 2019-09-20 at 3 07 10 PM

lists

Fixes #818 

A few things to call out here.

We use a new index view. That way this work doesn't get conflated with other views that use the hierarchy partial sets. Also per the discussion on 9/19 I think we want these as lists and not h elements.

TODOs:
 - [x] Figure out how to interleave the list elements
 - [x] Are these things collapsible?
      Yes but a future todo
 - [x] Spinning out and @jrgriffiniii is looking at this.. why isn't babel / es6 stuff working

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -34,14 +34,35 @@ class ContextNavigation {
       var $doc = $(resp);
       var newDocs = $doc.find('#documents')
         .find('article')
-        .toArray().map(el => new NavigationDocument(el).render());
+        .toArray().map(el => new NavigationDocument(el));
 
-      const docIndex = newDocs.findIndex(doc => doc.id === that.data.arclight.parentsParent)
-      console.log(docIndex, that.data.arclight);
-
+      const originalDocumentIndex = newDocs.findIndex(doc => doc.id === that.data.arclight.originalDocument);
       that.parentLi.find('.al-hierarchy-placeholder').remove();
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/arclight/context_navigation.js
```diff
@@ -0,0 +1,70 @@
...
+  render() {
+    return this.el.html();
+  }
+}
+
+
+Blacklight.onLoad(function () {
+
+  $('.context-navigator').each(function (i, e) {
+    const contextNavigation = new ContextNavigation(e);
```
jrgriffiniii: Integrating ES6 using `.es6` file extensions triggered the failures which I was encountering still, but it seems that just using `.js` functions without raising errors for me as well.

mejackreed: I've spun this issue out here: https://github.com/projectblacklight/arclight/issues/869

Issue #817, opened 2019-09-18
Components are not nesting properly in contents list. 


anarchivist:
See the attached screenshots - this impacts finding aids like the following in the demo site. It **should** look like the Alpha Omega Alpha finding aid.

<img width="1353" alt="Screen Shot 2019-09-18 at 10 38 20 AM" src="https://user-images.githubusercontent.com/73732/65172919-aedcb380-da02-11e9-94dd-f9edbb312f23.png">

I'm guessing that this is a case where `parent_ssm` (among other things?) might not be getting indexed properly; why Alpha Omega Alpha is working properly and others are not is less clear. See also @seanaery's issue #784. 

## Finding aids with issues

* [Alexander H. Stephens papers (Duke)](http://arclight-demo.projectblacklight.org/catalog/ahstephens)

<details><summary>(screenshots)</summary> 

**Existing presentation**

<img width="1353" alt="Screen Shot 2019-09-18 at 10 48 16 AM" src="https://user-images.githubusercontent.com/73732/65172701-55748480-da02-11e9-9383-26e2160bfd86.png">

**How Duke presents this**

<img width="1353" alt="Screen Shot 2019-09-18 at 10 48 54 AM" src="https://user-images.githubusercontent.com/73732/65172718-658c6400-da02-11e9-9f77-044e27e3e281.png">
</details>

* [Abdeen Jabara papers](https://arclight-demo.projectblacklight.org/catalog/umich-bhl-0234) - This one is particularly problematic for some reason... might be worth additional tickets.

<details><summary>(screenshots)</summary>

**Existing presentation**

<img width="1353" alt="Screen Shot 2019-09-18 at 10 46 14 AM" src="https://user-images.githubusercontent.com/73732/65172996-d7fd4400-da02-11e9-824b-da2d48892980.png">

**How Michigan presents this**

<img width="1353" alt="Screen Shot 2019-09-18 at 10 46 51 AM" src="https://user-images.githubusercontent.com/73732/65173067-f2372200-da02-11e9-9cd7-6614616284e0.png">
</details>

PR #864, opened 2019-09-20
Fixes #852 by properly accounting for numbered component elements

### LINKED ISSUE #852: Indexing EADs should account for numbered <c> elements
In accordance with the EAD 2002 standard, Documents may contain numbered variants of the `<c>` element from https://www.loc.gov/ead/tglib/elements/c01.html to https://www.loc.gov/ead/tglib/elements/c12.html.  Currently, using these in order to assign a `component_level` value is not supported by Traject:

```ruby
to_field 'component_level_isim' do |record, accumulator|
  accumulator << 1 + record.ancestors.count { |node| node.name == 'c' }
end
```
(https://github.com/projectblacklight/arclight/blob/master/lib/arclight/traject/ead2_config.rb#L284)

mejackreed:


camillevilla:
Related to #817.

PR #865, opened 2019-09-20
Optimize to_field 'parent_unittitles_ssm'

billdueber:
Quick benchmarking showed `parent_unittitles_ssm` to be taking up a huge proportion of the indexing time. This refactor speeds things up some.

It has larger effects with more heavily-nested documents.

`bx rake arclight:index FILE=../data/ead/umich-bhl/umich-bhl-2014150.xml` (7.3MB)
* Master (as of 20190920 5pm): 950 seconds
* This branch: 550 seconds (52% of master branch time)

`bx rake arclight:index FILE=../data/ead/InU-ewv/VAD1751.xml` (11MB)
* Master: 243 seconds
* This branch:  179 seconds (74% of master branch time)

(umich-bhl-9677.xml is the file that took @mejackreed 6,000 seconds as per https://code4lib.slack.com/archives/C43V2KVTM/p1568145216166500, so I'm not sure what I'm doing differently -- this is just on my desktop Mac)

I'm 99% sure the refactor is producing identical data (checked by indexing with both methods against a few files) but if someone else could take a look that'd be great.

### DIFF
--- a/lib/arclight/traject/ead2_config.rb
+++ b/lib/arclight/traject/ead2_config.rb
@@ -311,11 +311,10 @@ compose 'components', ->(record, accumulator, _context) { accumulator.concat rec
 
     # other components
     if parent_ssm && components
-      ancestors = parent_ssm.drop(1).map{ |x| [x] }
-      accumulator.concat components.select { |c| ancestors.include? c['ref_ssi'] }.flat_map{ |c| c['normalized_title_ssm'] }
+      ancestors = parent_ssm.drop(1).map { |x| [x] }
+      accumulator.concat components.select { |c| ancestors.include? c['ref_ssi'] }.flat_map { |c| c['normalized_title_ssm'] }
     end
   end
-
 
   to_field 'parent_unittitles_teim' do |_record, accumulator, context|
     accumulator.concat context.output_hash['parent_unittitles_ssm']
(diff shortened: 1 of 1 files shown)


mejackreed:
For better context, that 6000 seconds was on an AWS ec2 medium instance, with Solr likely not optimized.

lib/arclight/traject/ead2_config.rb
```diff
@@ -303,14 +303,16 @@
...
-      accumulator.concat Array
-        .wrap(context.clipboard[:parent].output_hash['components'])
-        .select { |c| c['ref_ssi'] == [id] }.map { |c| c['normalized_title_ssm'] }.flatten
+    parent_ssm = context.output_hash['parent_ssm']
+    components = context.clipboard[:parent].output_hash['components']
+
+    # other components
+    if parent_ssm && components
+      ancestors = parent_ssm.drop(1).map { |x| [x] }
+      accumulator.concat components.select { |c| ancestors.include? c['ref_ssi'] }.flat_map { |c| c['normalized_title_ssm'] }
```
jrgriffiniii: 👍 thank you for this, these lines seem to be identical to me.

I am curious as to why the speed was improved - perhaps because `#flatten` was originall invoked after the results of `.select { |c| c['ref_ssi'] == [id] }.map` were issued?  I can't imagine that introducing `parent_ssm.drop(1).map { |x| [x] }` could have changed much.

PR #873, opened 2019-09-24
Add sr-only label to bookmark icon

### LINKED ISSUE #862: Restore a perceivable label to Bookmark checkbox
Part of #820

Now that we're using the bookmark icon `svg` in place of the text label, the checkbox is effectively unlabelled for a screen reader. 

Instead of replacing the content of `<span>Bookmark</span>` with the `svg` do something like this:

```
<label>
   <input>
   <span><svg></span>
   <span class="sr-only">Bookmark</span>
</label>
```

- Retain the original text label but make it visible to SR-only when we want the icon alone
- Add the icon before the `sr-only` span

estelendur:
Fixes #862.

This works but it is ugly. Given the form_tag context, I'm not entirely sure there's a prettier way to do it, but I imagine there is something I don't know that could help.

### DIFF
--- a/app/views/catalog/_arclight_bookmark_control.html.erb
+++ b/app/views/catalog/_arclight_bookmark_control.html.erb
@@ -10,8 +10,8 @@
                  class: 'bookmark-toggle col-auto',
                  data: {
                    'doc-id' => document.id,
-                   present: blacklight_icon(:bookmark),
-                   absent: blacklight_icon(:bookmark),
+                   present: "#{blacklight_icon(:bookmark)}<span class=\"sr-only\">#{t('blacklight.search.bookmarks.present')}",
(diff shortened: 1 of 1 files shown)


mejackreed:
If we implement this: https://github.com/projectblacklight/blacklight/issues/2150

I think we could use those patterns which may allow us to not have to have the `sr-only` spans? What do you think @estelendur

estelendur:
@mejackreed That looks like it would do the trick, indeed.

labradford:
@mejackreed @estelendur does that mean hold off on merging this PR or does it mean go ahead and merge but we will need to revisit later?

estelendur:
@labradford @mejackreed I'm inclined to go ahead and merge this PR unless we want to push to get the blacklight change done this sprint.

mejackreed:
That works for me 👍

Issue #836, opened 2019-09-18
Suppress "collection_count"


jvine:
Part of #820

In some contexts, the search results display `<div class='al-collection-count'>` above the results. 

- It's redundant with the results count that's already present on the page
- It looks sort of like a heading, but semantically it's just text that's not really connected to anything
- It doesn't work correctly with the "group by collection" functionality (when a result set of collections is grouped by collection - the count goes to 0)

I propose suppressing / removing this section rather than trying to make it work functionally/semantically - it doesn't really add value to the page.

mejackreed:
Fixed in #875

Issue #820, opened 2019-09-18
Accessibility improvements: Headings, Landmarks, etc.


jvine:
This will be an inventory ticket for some broad (and some detailed) accessibility improvements. The overall goal is to ensure that:
- there is a meaningful heading structure that communicates the structure and focus of the page; with no missing heading levels
- there is a meaningful landmark structure that both passes audits (does not currently) and communicates the structure of the page
- svgs/icons are perceivable but don't get in the way
- some hidden pre-BS4 workarounds are updated or removed so they don't cause confusion.

Some of these are Blacklight-level changes, but they should be uncontroversial.

## Headings
- [x] Change `<h1>Archival Collections at Institution</h1>` to a `<span>` on all but the home page. #821
- [x] Add a page-specific `<h1>` to all page templates #834 
- [x] Make the `<title>` content match (or close to) the `<h1>` for the page #842
- [x] Update heading levels on the Repository page to eliminate missed levels #835 
- [x] Suppress "# collections" text on the Repository pages #836
- [x] Add/update headings in the tab sections on the Collection/Component pages #837
- [ ] Address the **`<h3>` everywhere** issue with component titles in hierarchy views

## Landmarks & roles
- [x] Make `<div id="main-container">` the `main` landmark #838
- [x] Add aria-labels to `<section>`s #841
- [x] Make the breadcrumb a labelled `<nav>` #839
- [ ] Add an aria-label to the search form [blacklight#2147](https://github.com/projectblacklight/blacklight/issues/2147)

## Jump-to links
- [ ] Add jump links to all pages #840

## svgs
- [ ] Make svg icons perceivable by screen readers #866
- [ ] Tooltips to be addressed at some future point

## Cleanup
- [ ] Move unassociated `sr-only` text to `aria-label` on results view button group
- [x] Hide pseudo-content (», others?) from SR. #854
- [x] Restore label to Bookmark checkboxes #862

more to come

seanaery:
Resolving this issue as part of a review of `arclight 1.0.x`. As of v1, aXe DevTools no longer reports any issues related to headings and landmarks. Any specific accessibility issues that remain should be submitted as individual tickets (if they haven't already been).

PR #875, opened 2019-09-24
Suppress collection_count #836

gordonleacock:
I hope this isn't overkill, I removed:
app/views/catalog/_collection_count.html.erb
spec/views/_collection_count.html.erb_spec.rb

And cleaned up all the specs accordingly.

From the ticket:
Part of #820

In some contexts, the search results display <div class='al-collection-count'> above the results.

It's redundant with the results count that's already present on the page
It looks sort of like a heading, but semantically it's just text that's not really connected to anything
It doesn't work correctly with the "group by collection" functionality (when a result set of collections is grouped by collection - the count goes to 0)
I propose suppressing / removing this section rather than trying to make it work functionally/semantically - it doesn't really add value to the page.


spec/features/collection_search_results_spec.rb
```diff
@@ -13,7 +13,7 @@
     it 'is active when collection search is activated' do
       visit search_catalog_path f: { level_sim: ['Collection'] }, search_field: 'all_fields'
       within '#main-container' do
-        expect(page).to have_css '.al-collection-count', text: /\d+ collections/
```
mejackreed: I'm wondering if we can just delete this entire file now?

gordonleacock: Yeah, that probably a better idea. Will do!

gordonleacock: 1. Done!

PR #878, opened 2019-09-25
Add repository/container icon to search result item breadcrumbs #829

### LINKED ISSUE #829: Add repository/container icon to search result item breadcrumbs
This applies to:

- List view and compact view of **flat search results**, using the repository icon
- List view and compact view of **grouped search results**, using the folder icon

Goals are:

- [x] Prepend a repository/container icon to the breadcrumb line in search result items

- [x] Ensure that when the breadcrumb line wraps, subsequent lines are left-aligned with the first line text, rather than the left-edge of the icon

### Example in desktop viewport
Screen Shot 2019-09-18 at 2 45 14 PM

### Example in mobile viewport

Screen Shot 2019-09-18 at 2 52 55 PM

### Icons
For the non-grouped result views, this [Material icon](https://material.io/resources/icons/?search=accoun&icon=account_balance&style=sharp) (no attribution needed) is the one to use as the repository icon.

For the grouped results, we start the breadcrumb at a container level, so we need to use the container (folder) icon we're already using elsewhere in the app.

### Implementation hints
Here's how I implemented the solution shown in the screenshots (basically adding the `margin-left` properties). I'm not suggesting this is the best way to do this (are there drawbacks to using negative margins like this?). It seems to work but maybe there are better ways?

```
.breadcrumb-links.col {
    margin-left: 20px;
}

.breadcrumb-links.col svg {
    height: 1rem;
    width: 1rem;
    fill: #6c757d;
    margin-left: -20px;
}
```

(I also added vertical spacing between the scope note and the breadcrumb line, but I believe that will be taken care of in #826.)

gordonleacock:
Closes #829.

Per the issue adds icons for flat and grouped search results pages:
List view and compact view of flat search results, using the repository icon
List view and compact view of grouped search results, using the folder icon

See https://github.com/projectblacklight/arclight/issues/829

I suggest having @ggeisler review the added styles. There may be room for improvement there.


jrgriffiniii:
As stated in the PR message, I defer to ggeisler.  However, this looks excellent to me:

image

app/assets/stylesheets/arclight/modules/search_results.scss
```diff
@@ -150,6 +150,24 @@ article.document {
   color: $body-color;
 }
 
+.al-search-result-index-article .breadcrumb-links {
+  .col {
+  flex-basis: auto;
+  margin-left: -12px;
```
jrgriffiniii: I was directed to refer to the classes in https://getbootstrap.com/docs/4.3/utilities/spacing/#notation for some rules I've applied involving `margin-left`, but those may have been specific to my cases.

ggeisler:
Just reacting to the visual appearance, I agree with @jrgriffiniii that this looks great. I'm assuming the grouped results version looks similar (but uses the folder icon in place of the repository icon) and the wrapping looks similar at all breakpoints.

app/assets/stylesheets/arclight/modules/search_results.scss
gordonleacock: Thanks @jrgriffiniii, I'll try those classes.

gordonleacock: @jrgriffiniii, used those spacing classes allowed me to clean up my scss. 👍

PR #880, opened 2019-09-25
861 limit repository masthead

### LINKED ISSUE #861: Remove masthead from search results if not faceted by Repository
When you limit search results by **Repository**, you get the Repository masthead. 
https://arclight-demo.projectblacklight.org/?f%5Brepository_sim%5D%5B%5D=Stanford+University+Libraries.+Special+Collections+and+University+Archives&q=&search_field=all_fields

This is as intended. Yay!

When you limit search results by **Collection** (or anything else), you _might_ get... the Repository masthead. 
https://arclight-demo.projectblacklight.org/?f%5Bcollection_sim%5D%5B%5D=Philip+G.+Zimbardo+papers%2C+1953-2011&q=&search_field=all_fields

I can't tell why this is happening, but it is misleading/disorienting, and was not intended by the original designs. 

**Acceptance criteria**

- [ ] Display the Repository masthead _only_ for searches with a Repository facet explicitly selected.

estelendur:
Fixes #861.

The problem was caused by `facets_from_request`, which is a bit of blacklight magic.

The solution uses another piece of magic, namely that the request parameters are magically available.


app/helpers/arclight_helper.rb
```diff
@@ -167,11 +167,12 @@ def fields_have_content?(document, field_accessor)
   #
   # @return [Repository]
   def repository_faceted_on
-    return unless try(:search_state)
+    search_fields = params.fetch('f', nil)
```
cbeer: I think Blacklight's `facet_field_in_params?` helper might do the trick here?

And, retrieving the value might just be `Array(facet_params['repository_sim']).first`?

estelendur: I have the impression that checking that there is only one repository is deliberate, but I don't actually know if we need to handle the theoretical case of having two repositories in the facets, since as far as I know they are 'and' rather than 'or' when you filter by them.

PR #884, opened 2019-09-26
Add sort option to group search results by collection #676

### LINKED ISSUE #676: Add sort option to group search results by collection
Part of #667.

To enable the user to switch to a view of search results where the results are grouped by collection, we need to add an option to the current sort menu:

Screen Shot 2019-08-29 at 3 15 26 PM

- [ ] Add a divider after the last item in the current dropdown menu: `<div class="dropdown-divider"></div>`

- [ ] Add a new option with the label: `Group by collection`. When selected, this option should take the user to the grouped view of search results, which will be described in forthcoming tickets.

gordonleacock:
Closed #676 by PR #884
Adds a new drop down item to the search results sort widget that provides a group by collection action.

1. Adds  app/views/catalog_sort_widget.html.erb (copied from blacklight)

2. Modifies it from the blacklight original by adding a link labelled "Group by collection" with this code
```<hr>
    <% add_hash = { "group" => "true" } %>
    <% new_params = params.permit(params.keys).merge(add_hash) %>
    <%= link_to t('arclight.search.sort_option_by_group'), new_params, { class: "dropdown-item" } %>```

3. Modifies spec/search_results_spec.rb to look for that new drop down item.


gordonleacock:
Wondered if using `params.permit(params.keys)` was a security problem too.

PR #885, opened 2019-09-27
add sr-only h2 to all tab content, use localizations to match tab name

### LINKED ISSUE #837: Add/update headings in the tab sections on the Collection/Component pages
Part of #820

Currently, the Overview and Contents tab have hidden `<h2>`s that use deprecated wording. The other tabs have no headings.

- [ ] Each tab panel should have an `<h2 class="sr-only">` as the first item in the panel (except Overview, which is fine where it is). 

- [ ] Each heading should use the same text as its parent tab, in a way that ensures they will remain in sync as the visible tab label is changed or translated.

- [ ] When the tab panel is empty (e.g. the tab is present but says "No online content"), the `<h2>` should not be present.

labradford:
closes #837 
Part of #820

### Collection and Component show pages:
Each tab panel should have an 'h2 class="sr-only"' as the first item in the panel (except Overview/(Collection context), which is fine where it is).

Each heading should use the same text as its parent tab, in a way that ensures they will remain in sync as the visible tab label is changed or translated.

When the tab panel is empty (e.g. the tab is present but says "No online content"), the 'h2' should not be present.


app/views/catalog/_collection_contents.html.erb
```diff
@@ -1,6 +1,6 @@
 <% document ||= @document %>
 
-<h2 class="sr-only">Collection inventory</h2>
+<h2 class="sr-only" data-sr-enable-me='true' data-has-contents=<%= t 'arclight.views.show.has_content' %>></h2>
```
jrgriffiniii: 👍

Issue #761, opened 2019-09-11
Format @render attributes for display (bold, italic, etc.)


noahgh221:
As of 9/11, the demo app is not applying any formatting based on values of the `@render`, which may appear as an attribute of the `<title>`, `<emph>` or `<titleproper>` elements.

From the EAD2002 Tag Library, here are possible values of the render attribute:

RENDER -- The formatting of the content of an element is controlled for display and print purposes. Available in <emph>, <title>, and <titleproper>. See also ALTRENDER. Values are:

altrender
bold
doublequote
bolddoublequote
bolditalic
boldsinglequote
boldsmcaps
boldunderline
italic
nonproport
singlequote
smcaps
sub
super
underline

Here is an example fixture that contains `<title @render="italic">Daily and Weekly Sun</title>`: https://arclight-demo.projectblacklight.org/catalog/ahstephens#background

Current display in demo app:
![image](https://user-images.githubusercontent.com/7328518/64720734-26599280-d499-11e9-9e28-9b2f060c38c0.png)

seanaery:
Looks like we'll probably need to build the equivalent of this (from SolrEAD) using Traject:
https://github.com/awead/solr_ead/blob/master/lib/solr_ead/formatting.rb

jlhardes:
I know this is closed now but this change might be what has created the problem reported in #990  - Title element not showing unless render attribute included

Issue #760, opened 2019-09-11
Format chronlist element set for display


noahgh221:
As of 9/11, the Demo application is not applying any formatting to the  `<chronlist>` element set that may appear as a child of a note field (most commonly `<bioghist>`, etc.). The `<chronlist>` element set is typically used to format biographical timelines.

The demo app is displaying the text of the elements as one text block, but not supplying any special formatting (e.g. bullets, newlines, etc.).

Example (see Biographical/Historical section): https://arclight-demo.projectblacklight.org/catalog/rushbenjaminandjulia

Current display:
![image](https://user-images.githubusercontent.com/7328518/64719983-661f7a80-d497-11e9-9e8d-887259d83f46.png)

EAD2002 Tag Library documentation on `<chronlist>` element: http://www.loc.gov/ead/tglib/elements/chronlist.html

For reference, here is how we (Duke) currently display `<chronlist>` elements within `<bioghist>`  for the collection referenced above: https://library.duke.edu/rubenstein/findingaids/rushbenjaminandjulia/#historicalnote

![image](https://user-images.githubusercontent.com/7328518/64720032-83544900-d497-11e9-80be-9ddc760f212b.png)

CC @seanaery

ggeisler:
Below is a recommendation for how to deal with the `<chronlist>` element. As @noahgh221 says above, we don't currently do anything with this element but it should be fairly straightforward to work with it, since there are only a few child elements of `<chronlist>`. Thanks to @noahgh221 for the relevant pointers.

### How a chronlist should look
Here's the target for displaying the chronlist (note there are two separate chronlists displayed here, one for Benjamin and one for Julia):

<img width="1011" alt="Screen Shot 2019-09-20 at 4 43 30 PM" src="https://user-images.githubusercontent.com/101482/65364304-d29c2700-dbc5-11e9-9563-0aad85c812d4.png">

---

### Chronlist elements mapping
Below is an annotated screenshot showing how the `<chronlist>` child elements are mapped to HTML. For reference, here's the [relevant page in the EAD 2002 spec](https://www.loc.gov/ead/tglib/elements/chronlist.html). Some details:

* The `<chronlist>` structure is mapped to an HTML table.
* If there is a `<head>` or `<listhead>` element within the chronlist, it is mapped to a `<caption>`.
* The table gets two `<th>` elements to set up table column headers of Date and Event (see example code for the table below).
* `<chronitem>` becomes a `<tr>` of the table.
[...]

PR #888, opened 2019-09-27
Convert EAD formatting for lists & text to HTML tags via a Traject ma…

### LINKED ISSUE #758: Format list/item elements in note fields for display
As of 9/11, the Demo application is not properly formatting `<list>` and `<item>` elements that appear as children in note fields (`<scopecontent>`, `<bioghist>`, etc.).  

The demo app is displaying the text of the elements as one text block, but not supplying any special formatting (e.g. bullets, numbers, newlines, etc.).

Example: https://arclight-demo.projectblacklight.org/catalog/brownfrankclydeaspace_ccacfdefd3fa0c00b03ef491d0a85bd0

EAD Tag Library documentation on where `<list>` elements may appear: https://www.loc.gov/ead/tglib/elements/list.html

For reference, here is how we (Duke) display lists for the same component referenced above: https://library.duke.edu/rubenstein/findingaids/brownfrankclyde/#aspace_ccacfdefd3fa0c00b03ef491d0a85bd0

CC @seanaery

seanaery:
…cro w/XSLT; render HTML for some fields. Closes #758, #760, #761

### Summary of Changes

- Adds a `format_tags` macro that can be easily added to Traject `to_field` indexing rules for fields where we'd expect to find formatting tags
- Uses an XSLT file with templates to convert formatting-related EAD XML to corresponding HTML tags during indexing
- Defines HTML transform rules for: `<list>`, `<deflist>`, `<chronlist>` and several render attributes e.g., `<emph render="bold" />`
- Adds a `render_html_tags` helper method that, if used in a field config, will use `.html_safe` to render the value as html.
- Configures all searchable note fields (`<abstract>`, `<scopecontent>`, `<bioghist>`, etc.) to index & display with these formatting rules applied (on collection and component level).

### Screenshots
Screen Shot 2019-09-27 at 2 02 28 PM

Screen Shot 2019-09-27 at 2 04 52 PM

### DIFF
--- a/lib/arclight/traject/ead_formatting.xsl
+++ b/lib/arclight/traject/ead_formatting.xsl
@@ -1,0 +1,188 @@
+<?xml version="1.0" ?>
+<xsl:stylesheet version="1.0"
+     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:ead="urn:isbn:1-931666-22-9"
+    xmlns:xlink="http://www.w3.org/1999/xlink"
+    exclude-result-prefixes="ead xlink"
+>
+
+  <xsl:output omit-xml-declaration="yes" encoding="UTF-8" method="html" />
+
+  <xsl:template match="/">
+      <xsl:apply-templates />
+  </xsl:template>
+
+  <!--
+  ~~~~~~~~~~~~~~~~~~~~~~~~
+  Lists
+  ~~~~~~~~~~~~~~~~~~~~~~~~
+  -->
+
+  <xsl:template match="list[@type='ordered']">
+    <xsl:apply-templates select="head" />
+    <ol><xsl:apply-templates select="item" /></ol>
+  </xsl:template>
+
+  <xsl:template match="list[@type='deflist']">
+    <table class="table deflist">
+      <xsl:apply-templates />
+    </table>
+  </xsl:template>
+
+  <xsl:template match="list[@type='deflist']/listhead">
+    <thead>
+      <tr>
+        <th scope="col">
+          <xsl:apply-templates select="head01" />
+        </th>
+        <th scope="col">
+          <xsl:apply-templates select="head02" />
+        </th>
+      </tr>
+    </thead>
+  </xsl:template>
+
+  <xsl:template match="list">
+    <xsl:apply-templates select="head"/>
+    <ul><xsl:apply-templates select="item" /></ul>
+  </xsl:template>
+
+  <xsl:template match="defitem">
+    <tr>
+      <td><xsl:apply-templates select="label" /></td>
+      <td><xsl:apply-templates select="item" /></td>
+    </tr>
+  </xsl:template>
+
+  <xsl:template match="list/head">
+    <strong><xsl:apply-templates /></strong>
+  </xsl:template>
+
+  <xsl:template match="list/item">
+    <li><xsl:apply-templates /></li>
+  </xsl:template>
+
+  <!--
+  ~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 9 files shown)


seanaery:
A couple things I'm not 100% confident about:

1. This module / macro / mixin setup works but I don't know if I have it set up and named in the most idiomatic way, and/or if there's a better way to make this play nicely with Traject and TrajectPlus. My experience here is limited, so I especially welcome feedback on this part.

2. Is it acceptable to use XSLT like this for these transformations? I explored some alternatives like [Nokogiri methods](https://github.com/sparklemotion/nokogiri/wiki/Cheat-sheet) to transform the tags, but found XSLT to be more intuitive, especially for cascading template calls where formatted elements are nested in other formatted elements.

lib/arclight/traject/ead_formatting.xsl
```diff
@@ -0,0 +1,188 @@
+<?xml version="1.0" ?>
+<xsl:stylesheet version="1.0"
```
cbeer: Is this an XSLT of our own creation, or some standard one?

seanaery: @cbeer Good question; this is one I created. I don't believe there's any _standard_ EAD->HTML XSLT, though I have at least been able to incorporate some transformation ideas from our (Duke's) legacy finding aids applications, and from [ArchivesSpace source code](https://github.com/archivesspace/archivesspace/blob/master/stylesheets/as-ead-html.xsl).

cbeer:
> Is it acceptable to use XSLT like this for these transformations?

I'd be a little more comfortable (since the XSLT is original work and not something standardized) if we performed the EAD -> HTML transformation in ruby-land, and I'd strongly suggest doing it at render time in order to give downstream applications the easiest opportunity to customize the transformation with the least amount of work. Rails has a built-in HTML sanitization system built on loofa which may be useful, maybe something like:

```
def render_html_tags(args)
my_custom_scrubber =  Loofah::Scrubber.new do |node|
  node.name = 'strong' if node.attr('render') == 'bold'
  node.name = 'sub' if node.attr('render') == 'sub'
  node.name = 'ul' if node.attr('render') == 'list'
  # ...
  node.scrub!(Rails::Html::PermitScrubber.new)
end

safe_join(args[:value].map { |value| content_tag(:p, sanitize(value, scrubber: my_custom_scrubber)) })
end
```

seanaery:
> I'd be a little more comfortable (since the XSLT is original work and not something standardized) if we performed the EAD -> HTML transformation in ruby-land, [...]

Thanks @cbeer for this perspective. I had explored this non-XSLT approach a bit but got stuck, though your helpful code snippet here may be just what I needed to get unstuck.  I hadn't used `Loofah::Scrubber`before but will see if I can get that to replicate the logic from the XSLT. Extra sanitizing seems like a nice bonus too. 

> and I'd strongly suggest doing it at render time in order to give downstream applications the easiest opportunity to customize the transformation with the least amount of work.

This sounds reasonable. So in that case, we would just use `to_text: false` in the Traject config (using no additional macro) for formatted fields and retain all the esoteric EAD markup in those fields' values in Solr, right?

escowles:
👍 to avoiding XSLT here if at all possible.  I don't know if we're typical, but we have very, very few developers who are competent in XSLT and there is virtually no overlap with our Ruby developers.

seanaery:
Thanks @escowles -- that's helpful to know. I'll plan to rework this in the next sprint to go XSLT-free.

seanaery:
I'll close out this PR with the intention of replacing it with a new one in the coming days that'll use a different strategy for transforming the tags.

Issue #811, opened 2019-09-18
Update display of current component and siblings in collection context


camillevilla:
Split out of #670 / #673

   - [x] Current component is highlighted with a yellow background (1, below)
   - [x] Because the current component may be far down in the tree, portions of the tree are hidden to ensure the current component is visible in the tab. 
   - [x] The entire remaining tree _below_ the current component is shown. 

![](https://user-images.githubusercontent.com/6945691/63970094-29479280-ca58-11e9-96d0-adbffe79733b.png)

mejackreed:
Resolved in #889 (as far as I can tell). Please open a new issue for a missing piece of this. cc @jrgriffiniii

Issue #457, opened 2017-06-14
Collection-level online content tab displaying only 10 components


anarchivist:
The "Online content" tab on a collection detail page displays components with digital objects when there are no collection-level digital objects. Currently, it is displaying only 10 components. 

**Acceptance criteria (for discussion/negotiation)**

* When on the Online content tab on a collection detail page and there are no collection-level display objects, display all of the components with digital objects.
* Consider providing some pagination on this display.

**Screenshot**

![screen shot 2017-06-14 at 2 43 16 pm](https://user-images.githubusercontent.com/73732/27156117-8b74bd66-5110-11e7-80b6-23ca914de418.png)

jrgriffiniii:
I'm still working to make progress on this.  The approach I have been taking is to add the `results_pagination` partial into https://github.com/projectblacklight/arclight/blob/master/app/views/catalog/_index_online_contents_default.html.erb (please see https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_search_results.html.erb#L28), but this required debugging in Kaminari: https://github.com/kaminari/kaminari/blob/master/kaminari-core/lib/kaminari/helpers/helper_methods.rb#L111

jrgriffiniii:
The approach I was taking was an erroneous one.  I have introduced JavaScript changes in https://github.com/projectblacklight/arclight/compare/master...jrgriffiniii:issues-457-jrgriffiniii-online-content-results which provides for pagination in the AJAX-loaded collection navigation element.

However, I did not have fixtures for 11 documents linking to online content, and hence I just copied spec/fixtures/ead/sul-spec/a0011.xml using different `eadid` values - is this acceptable for a pull request, or are there better fixtures which should be used?

mejackreed:
@jrgriffiniii Feel free to push up a WIP PR with this.

Is there a way to isolate the JavaScript changes to the individual pieces that enable the pagination / more results? I also see a bit of a refactor in there and trying to determine which pieces we need.

WRT to the fixtures. I think what you have done is ok. But we could also edit the current EAD fixtures to add the additional components in which may be a bit more surgical for this feature. Happy to pair up w/ you to do that.

@anarchivist is there a real world EAD in the wild we should target for this?

anarchivist:
Here are a couple options from Stanford:

[...]

PR #889, opened 2019-09-27
Implements an "expand" button for cases where the Collection Context interface contains more than 1 previous sibling document for items within a Collection

jrgriffiniii:
Advances #811, please see the following screenshots for the updated interface:

image
image

Please note that I am using the fixture referenced in https://github.com/projectblacklight/arclight/issues/457#issuecomment-522094674.  Shall I include this, or might there be a preferred fixture for these cases?

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,6 @@
+---
+version: "2"
+exclude_patterns:
+  - "app/assets/javascripts/arclight/context_navigation.js"
+  - "vendor/assets/javascripts/**/*.js"
+  - "spec/**/*.rb"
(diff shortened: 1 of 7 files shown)


mejackreed:
@jrgriffiniii If you can get away without adding a new fixture, I think that that would be preferred. But if not, totally understandable.

jrgriffiniii:
If I am understanding the documentation correctly, the JavaScript complexity is analyzed using ESLint: https://docs.codeclimate.com/docs/eslint.  I can resolve this by offering the needed `.eslintrc` configuration settings.

jrgriffiniii:
Following a review by mejackreed, at least one bug was identified in this which must be addressed.

Issue #637, opened 2019-08-26
Ensure component title wraps appropriately


ggeisler:
Part of #629 / #628.

I'm not sure how much, if any, work this will require, but after issues #630 and #631 have been completed, we should ensure that the component title wraps as necessary while maintaining the desired vertical placement below the last breadcrumb node.

Here's what I mean: 

Ideally we'd like the component title to extend full-width of the page (to the right) whenever possible, such as this:

<img width="1146" alt="Screen Shot 2019-08-26 at 2 54 18 PM" src="https://user-images.githubusercontent.com/101482/63726294-bb118e80-c811-11e9-854e-5a93f8d8b551.png">

However, the priority is that the title fit into the breadcrumb hierarchy (visually) with no unusual vertical gap between the final node of the breadcrumbs and the title. In some cases, this means that a long title will not be able to extend full-width because it will be in the same horizontal space as the component context/actions box described in #631. For example, if the previous example was one level higher in the hierarchy, extending the title all the way to the right would conflict with the context/actions box:

<img width="986" alt="Screen Shot 2019-08-26 at 3 19 04 PM" src="https://user-images.githubusercontent.com/101482/63727386-ffeaf480-c814-11e9-8907-1a49ec94b8a3.png">

Rather than create a vertical gap above the title that would enable the title to extend full-width below the context/actions box, we'd prefer that it wraps at the box like this:

<img width="986" alt="Screen Shot 2019-08-26 at 2 56 02 PM" src="https://user-images.githubusercontent.com/101482/63726419-10e63680-c812-11e9-9079-2dd5322a3936.png">

So:

- [ ] Component title extends full-width of the page if there is no conflict with the context/actions box

- [ ] Component title wraps at the context/actions box if it would otherwise need to move down vertically to avoid overlapping with the context/actions box.

gordonleacock:
@ggeisler Hi Gary, the way we display the action box in PR##695 (still not completed) is by placing the breadcrumb in a col-md-9 div and the action box in col-md-3 div, so there is no overlap. This design is only for the components show page not the collections show page.

gordonleacock:
@ggeisler Open to meeting this requirement differently if col-md-9/col-md-3 solutions isn't adequate but would need styling guidance on show best to do that.

ggeisler:
@gordonleacock I think it is too soon to say whether this is closeable. I need to see the actions box in action, with the changes in other PRs related to the action box merged, before we can judge how well the component title wrapping works.

[...]

PR #896, opened 2019-10-01
Removes duplicate spacing classes and use aeon web ead button classes

mejackreed:
This is a follow on based on evaluation of #640 . There seemed to be some duplicate classes here that interacted and the aim here was to overall simplify the approach for styling this area.

@ggeisler mind weighing in here?

## Before
Screen Shot 2019-10-01 at 1 25 54 PM
## After
Screen Shot 2019-10-01 at 1 26 03 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -16,6 +16,17 @@
   font-size: $font-size-sm;
   height: 100%;
   padding: $spacer;
+  
+  :first-child {
+    padding-top: 0;
+  }
+  :last-child {
+    padding-bottom: 0;
+  }
+
+  > div {
+    padding: ($spacer / 3) 0;
+  }
 
   &-options {
     display: flex;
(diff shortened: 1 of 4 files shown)


ggeisler:
@mejackreed Yep, this looks better and I'm fine with the spacing in general (the bottom spacing seems a tad too much to me, but maybe that could be left local implementation fine-tuning, or I could work on tweaking it in a separate PR later). 

The remaining issue is with the fixed width of the box, which results in too much right spacing in many cases:

Screen Shot 2019-10-01 at 12 35 50 PM

Somewhere I think we discussed making the width of the box variable rather than fixed (which is why we get the unwanted right-side space now) but I can't find it at the moment. The goal is for the width of the box to just be wide enough to contain its contents with even spacing on all sides.

If you want I could make a separate ticket just focused on the variable-width aspect of things.

mejackreed:
👍  @ggeisler wrt the width.. I think that will need to be implemented with #637. It seems to be indicated there already.

ggeisler:
Okay sounds good.

PR #903, opened 2019-10-02
Allow actions box to be variable width and make the containers, and…

jkeck:
…buttons flow inline.

Not sure if this is ultimately better, but a few things that could potentially improve things that I found when investigating #637.

## Before
title-actions-responsive-before

## After
title-actions-responsive-after

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -15,6 +15,7 @@
   background-color: $gray-200;
   font-size: $font-size-sm;
   height: 100%;
+  min-width: 200px;
   padding: $spacer;
   
   :first-child {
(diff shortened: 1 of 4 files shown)


ggeisler:
Thanks for working on this @jkeck. There isn't a dramatic improvement in your example, where all elements of the actions box are present, but I do think what you've done might make things look better when there are fewer elements in the actions box, such as when the collection/container info is not available to display.

Plus your solution seems to fix the right alignment issue where previously the right edge of the actions box seemed to extend slightly further right than the right edge of the tabbed container area below it. They seem to line up correctly now,

There are also some spacing tweaks I think we can make (I'd suggest not worrying about them for this PR, though) that can tighten up the box in the mobile viewports and make those cases look a bit better.

So I'd be happy going with this, from a UI perspective.

PR #904, opened 2019-10-02
Bugfixes for ID-less components: prepend al_ to IDs & correctly set a…

### LINKED ISSUE #895: Component nesting not working in context tree when ID-less
Noted by @cmkpowell today, there's an indexing bug that is making ID-less components not display/behave correctly in the Contents or Context tabs in the UI.

**Example:**
Screen Shot 2019-10-01 at 3 12 03 PM

**Example Page:**
https://arclight-demo.projectblacklight.org/catalog/umich-wcl-M-2370mac

**Data:**
https://github.com/sul-dlss/arclight-demo/blob/master/data/ead/umich-wlcl/macarthur_final.xml

**Legacy UI:**
https://quod.lib.umich.edu/c/clementsead/umich-wcl-M-2370mac?rgn=Entire+Finding+Aid;view=text;q1=macarthur#Contents%20List

I believe the `parent_ssm` field is not capturing the array of parent IDs correctly [in our Traject config](https://github.com/projectblacklight/arclight/blob/master/lib/arclight/traject/ead2_config.rb#L305-L325) and there may be other related issues.

CC: @anarchivist

seanaery:
…ncestor trail. Fixes #895.

### Problems Solved (see #895):
- Components with missing IDs were not rendering with the correct nesting in the context tree nor in the breadcrumb area
- Many components with missing IDs would not expand/collapse their descendant nodes upon clicking +/- This was due to a limitation of Bootstrap (collapsed section IDs have to begin with a letter)

### Solutions Implemented in this PR:
- Prepend minted component IDs with `al_` so they begin with a letter (fixes the Bootstrap collapse/expand)
- Write the minted `id` back to the node's `id` attribute during indexing so it gets subsequently picked up correctly in descendant nodes' ancestor trail array (in `parent_ssm`).

Screen Shot 2019-10-02 at 12 57 28 PM

Screen Shot 2019-10-02 at 12 57 07 PM

### DIFF
--- a/lib/arclight/hash_absolute_xpath.rb
+++ b/lib/arclight/hash_absolute_xpath.rb
@@ -24,7 +24,7 @@ module Arclight
     end
 
     def to_hexdigest
-      self.class.hash_algorithm.hexdigest(absolute_xpath)
+      self.class.hash_algorithm.hexdigest(absolute_xpath).prepend('al_')
     end
 
     def absolute_xpath
--- a/lib/arclight/traject/ead2_config.rb
+++ b/lib/arclight/traject/ead2_config.rb
@@ -252,6 +252,7 @@ compose 'components', ->(record, accumulator, _context) { accumulator.concat rec
                          "The ID of this document will be #{parent_id}#{hexdigest}."
                        ].join(' ')
                      end
+                     record['id'] = hexdigest
                      hexdigest
                    else
                      record.attribute('id')&.value&.strip&.gsub('.', '-')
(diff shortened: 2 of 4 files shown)


lib/arclight/hash_absolute_xpath.rb
```diff
@@ -24,7 +24,7 @@ def initialize(node)
     end
 
     def to_hexdigest
-      self.class.hash_algorithm.hexdigest(absolute_xpath)
+      self.class.hash_algorithm.hexdigest(absolute_xpath).prepend('al_')
```
mejackreed: Could we re-verify this again? I was able to get this working without prepending `al`. Could we chat quickly about this?

mejackreed: Thanks for pairing up @seanaery on this. I was able to see t he issue in the "Contents" collapse and expand and was confused.

PR #915, opened 2019-10-03
update context page to match search results

### LINKED ISSUE #908: Collection context: Update layout to have look and feel of search results
Similar to other search results, collection context should have the following:

See 4, 5, 6

 - [ ] style and layout of search results
   - [ ] online content icon (5)
   -  [ ] bookmark icon (6)
 - [ ] container info (4)

labradford:
Closes #908 

### Issue
update collection context page with style and layout of search results
add icon to title
add online content icon
add bookmark icon and checkbox
add container info

### Discussion
The nesting of the child components on mobile will probably need design review. If there are several layers of nesting, the indenting might be an issue on mobile.

#### Large screen:
desktop-contents

#### Mobile screen:
mobile-contents

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -148,6 +148,7 @@ class ContextNavigation {
       that.parentLi.after(renderedAfterDocs).fadeIn(500);
     }
     that.truncateItems();
+    Blacklight.doBookmarkToggleBehavior();
   }
 
   // eslint-disable-next-line class-methods-use-this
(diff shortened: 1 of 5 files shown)


ggeisler:
@labradford Regarding the indentation on mobile, that's definitely a concern we've discussed. One idea @jvine mentioned but that I'm not sure was formally proposed is to reduce the level indention by half on mobile. Another thing that might help a tiny bit is, as part of the visual design work I'm doing, I was thinking about reducing the spacing between the container icon and the item title.

But it might make sense to leave those issues for later because we (@jvine and I) might be able to test out potential ideas for reducing the indentation better when we can work with the updates in this PR directly.

labradford:
@ggeisler I think once this is on the demo site where there are more levels, it will be easier to see if this is a problem or not.

PR #916, opened 2019-10-04
Add a drop down to allow users to search within the current collection…

### LINKED ISSUE #719: Add a "scope" dropdown to the search bar
Part of #716 

- [x] **Add a dropdown to the search bar, in front of the form.**
   - dropdown uses the same button+menu structure as the "Sort" dropdown
   - add `btn-lg` class
   - button text is "Search" + user's / default selection
   - selections are:
      - all collections
      - this collection

Screen Shot 2019-09-04 at 2 34 44 PM

- [x] **If there is no current collection selected (*), disable "this collection"**
Screen Shot 2019-09-04 at 2 43 24 PM

- [x] **If there is a current collection selected (*) make "this collection" the default**
Screen Shot 2019-09-04 at 2 43 56 PM

### * "Current collection selected"

If the user is **on any component page**, searching within the collection that component belongs to should be the default behaviour. 

If the user is **in a search result with a collection limit applied**, they are de facto searching within "this collection" until the collection limit is removed. I think the search box should reflect that: so the default would be "this collection".

**That implies that selecting "all collections" in the dropdown _would also_ remove the collection limit.**

jkeck:
…context or all collections.

Closes #719 

Screen Shot 2019-10-04 at 11 16 20 AM

### DIFF
--- a/app/models/concerns/arclight/catalog.rb
+++ b/app/models/concerns/arclight/catalog.rb
@@ -4,6 +4,17 @@ module Arclight
   ##
   # Arclight specific methods for the Catalog
   module Catalog
+    extend ActiveSupport::Concern
+
+    included do
+      before_action only: :index do
+        if (params.dig(:f, :collection_sim) || []).any?(&:blank?)
+          params[:f][:collection_sim].delete_if(&:blank?)
+          params[:f].delete(:collection_sim) if params[:f][:collection_sim].blank?
+        end
+      end
+    end
+
     ##
     # Overriding the Blacklight method so that the hierarchy view does not start
     # a new search session
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -1,0 +1,34 @@
+<%
+# Overridden from Blacklight to add a drop down that allows the user to choose to search w/i the collection or all collections
+%>
+<%= form_tag search_action_url, method: :get, class: 'search-query-form', role: 'search' do %>
+  <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:q, :search_field, :qt, :page, :utf8).merge(f: (search_state.params_for_search[:f] || {}).except(:collection_sim))) %>
+  <div class="d-md-flex">
+    <%= render 'catalog/within_collection_dropdown' %>
+
+    <% if search_fields.length > 1 %>
+      <label for="search_field" class="sr-only"><%= t('blacklight.search.form.search_field.label') %></label>
+    <% end %>
+    <div class="input-group">
+      <% if search_fields.length > 1 %>
+          <%= select_tag(:search_field,
+                         options_for_select(search_fields, h(params[:search_field])),
+                         title: t('blacklight.search.form.search_field.title'),
+                         id: "search_field",
+                         class: "custom-select search-field") %>
+      <% elsif search_fields.length == 1 %>
+        <%= hidden_field_tag :search_field, search_fields.first.last %>
+      <% end %>
+
+      <label for="q" class="sr-only"><%= t('blacklight.search.form.search.label') %></label>
+      <%= text_field_tag :q, params[:q], placeholder: t('blacklight.search.form.search.placeholder'), class: "search-q q rounded-0 form-control", id: "q", autofocus: presenter.autofocus?, data: { autocomplete_enabled: presenter.autocomplete_enabled?, autocomplete_path: search_action_path(action: :suggest) }  %>
+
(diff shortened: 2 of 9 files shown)


app/models/concerns/arclight/catalog.rb
```diff
@@ -4,6 +4,18 @@ module Arclight
...
   # Arclight specific methods for the Catalog
   module Catalog
+    extend ActiveSupport::Concern
+
+    included do
+      before_action only: :index do
+        if params.dig(:f, :collection_sim) && params[:within_collection] == 'false'
+          params[:f].delete(:collection_sim)
+          params.delete(:within_collection)
+          redirect_to url_for(params.to_unsafe_h)
```
jkeck: Is there a better way to do this?

app/views/catalog/_search_form.html.erb
```diff
@@ -0,0 +1,39 @@
...
+# and add a drop down to allow the user to choose to search w/i the collection or all collections
+%>
+<%= form_tag search_action_url, method: :get, class: 'search-query-form', role: 'search' do %>
+  <% if results_view? %>
+    <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:q, :search_field, :qt, :page, :utf8, :within_collection)) %>
+  <% elsif record_view? %>
+    <%= render_hash_as_hidden_fields({ f: { collection_sim: [@document.collection_name] } }) %>
+  <% end %>
+  <div class="d-md-flex">
+    <%= render 'within_collection_dropdown' if record_view? || (results_view? && has_search_parameters?) %>
```
jkeck: I'm assuming that we only render the drop down in search results and the record view.  The search form also shows up on the home page and in the repositories view.  Should it be in all those places too?

app/views/catalog/_within_collection_dropdown.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="input-group w-50">
+  <span class="input-group-prepend">
+    <label class="input-group-text" for="within_collection">
+      <%= t('arclight.within_collection_dropdown.label') %>
```
jkeck: @jvine the current label is "Search" and wasn't sure if that is what we would ultimately want here.  We can add screen ready only text so the label is visually "Search" but reads as "Search...something else" if that would be preferable.  Wasn't exactly sure.

app/models/concerns/arclight/catalog.rb
cbeer: What if we gave the dropdown a little more smarts (probably better than this, but..)?

```
...
  <select
    ...
    name="f[collection_sim][]"
  >
    <option value="">
      ...
    </option>
    <option
      value="<%= Array(params[:f][:collection_sim]).first || @document && @document.collection_name %>">
     ...
    </option>
  </select>
...
```

cbeer: Oh, and always exclude it from the hidden fields:

```erb
    <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:q, :search_field, :qt, :page, :utf8, :f).merge(f: search_state.params_for_search[:f].except(:collection_sim))) %>
```

cbeer: And, I guess, we still need something to prune empty values from the params.. ugh.

jkeck: Yeah, it appears that we would, and also need to redirect unless we want to carry around an empty collection_sim facet value in the params (and just ignore it).  I'm indifferent about the approach.  One one hand, doing this doesn't add an additional parameter to the URL but on the other seems to be a little less clean wrt to the facet parameters hanging around and in-view parameter manipulation.

I'm happy to stamp out this approach if it is deemed preferable.

app/views/catalog/_within_collection_dropdown.html.erb
jvine: I think something else already has the label of "Search" - this might be better as "Search within" where "within" is hidden.

app/views/catalog/_search_form.html.erb
jvine: It is meant to be present on all pages, even though it can only have one choice (all collections) at the home and repositories pages. It is intended to also support future options that may be available from those locations, like "digital content".

app/models/concerns/arclight/catalog.rb
jkeck: @cbeer let me know if what I've done is in the spirit of what you were thinking (and if there is any potential issues w/ the ignored `f` param that I was trying to avoid w/ my redirect).

PR #919, opened 2019-10-07
Convert EAD formatting tags to HTML at render time. Closes #758,…

### LINKED ISSUE #758: Format list/item elements in note fields for display

seanaery:
### Summary of Changes

This is a rewrite of PR #888 with two major differences per discussion on that ticket: 1) Uses Ruby only for transformations (XSLT free); 2) Performs the transformation to HTML at render time rather than pre-indexing.

- Adds a `render_html_tags` helper method that transforms EAD XML tags stored in Solr fields into sanitized HTML
- Defines HTML transform rules for: `<list>`, `<deflist>`, `<chronlist>` and several render attributes e.g., `<emph render="bold" />`
- Configures all searchable note fields (`<abstract>`, `<scopecontent>`, `<bioghist>`, etc.) to index nodes as XML rather than text (`to_text: false`) and display with these formatting rules applied (on collection and component level).

### Screenshots
al-render
al-lists
al-deflists
al-chronlists

### DIFF
--- a/app/controllers/concerns/arclight/ead_format_helpers.rb
+++ b/app/controllers/concerns/arclight/ead_format_helpers.rb
@@ -1,0 +1,228 @@
+# frozen_string_literal: true
+
+module Arclight
+  ##
+  # A module to add EAD to HTML transformation rules for Arclight
+  module EadFormatHelpers # rubocop:disable Metrics/ModuleLength
+    extend ActiveSupport::Concern
+    include ActionView::Helpers::OutputSafetyHelper
+
+    def render_html_tags(args)
+      values = args[:value] || []
+      values.map! do |value|
+        html_sanitize(transform_ead_to_html(value))
+      end
+      values.map! { |value| wrap_in_paragraph(value) } if values.count > 1
+      safe_join(values.map(&:html_safe))
+    end
+
+    private
+
+    def transform_ead_to_html(value)
+      Loofah.xml_fragment(condense_whitespace(value))
+            .scrub!(ead_to_html_scrubber).to_html
+    end
+
+    def html_sanitize(value)
+      Loofah.fragment(value).scrub!(:strip).to_html
+    end
+
+    def ead_to_html_scrubber
+      Loofah::Scrubber.new do |node|
+        format_render_attributes(node) if node.attr('render').present?
+        format_lists(node) if %w[list chronlist].include? node.name
+        node
+      end
+    end
+
+    def condense_whitespace(str)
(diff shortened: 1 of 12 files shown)


seanaery:
Heads-up @cbeer -- I'm wondering if this PR adequately addresses concerns from previous PR #888. A couple potential drawbacks I can see from this approach (though these could be worthwhile tradeoffs?):

1. There's a lot of node-sculpting happening at render time, which could potentially impact performance.
2. Any other applications getting the Solr data (e.g., from Blacklight's raw JSON API) would also bear responsibility for transforming/sanitizing the tags. This would be less onerous if what we stored in Solr for these fields was the final cleaned/sanitized HTML rather than EAD.

app/controllers/concerns/arclight/ead_format_helpers.rb
```diff
@@ -0,0 +1,228 @@
...
+        html_sanitize(transform_ead_to_html(value))
+      end
+      values.map! { |value| wrap_in_paragraph(value) } if values.count > 1
+      safe_join(values.map(&:html_safe))
+    end
+
+    private
+
+    def transform_ead_to_html(value)
+      Loofah.xml_fragment(condense_whitespace(value))
```
cbeer: It looks like `.scrub!` is chainable; is there a reason we'd want to get the transformed-but-not-stripped HTML output? Or, could the `ead_to_html_scrubber` method take care of stripped all other elements?

camillevilla: Fair question, but I think @seanaery made some good progress here and I would be ok kicking this to a future refactor.

seanaery: @cbeer Here's the refactor: PR #952 ... Doesn't seem to break anything :-)

Issue #680, opened 2019-08-30
Show online content at every level of the hierarchy


jvine:
Part of #678 , related to #679 .

- [ ] If a component at any level of the hierarchy has children - direct or indirect - with online content, display an "Online content" tab listing those children as a search result.
- [ ] If a component at any level of the hierarchy has direct online content, display it above the tabs. 
- [ ] Both conditions can coexist. A component may have DAO links above the tabs, as well as an "Online content" tab with children listed.
- [ ] If a component has no online children, there is no "Online content" tab. (i.e. we are no longer showing a disabled tab that says "No online content")

**Collection** 
Shows all children with direct online content in the entire collection. 
<img width="1003" alt="collection online content" src="https://user-images.githubusercontent.com/6945691/63987798-5151e880-ca8e-11e9-9d00-06a14d7bd99c.png">

**Series** 
Shows all children with direct online content, scoped to this series. Includes all sub- and sub-sub-(sub-sub-)series. Count may be less than the collection-level list.
<img width="998" alt="series online content" src="https://user-images.githubusercontent.com/6945691/63987803-53b44280-ca8e-11e9-95eb-e943c2090e3a.png">

**Sub-series** 
Shows all children with direct online content, scoped to this sub-series (and below). Count may be less than the series list.
<img width="1001" alt="subseries online content" src="https://user-images.githubusercontent.com/6945691/63987807-56af3300-ca8e-11e9-851b-02876568d226.png">

**File** 
Item has direct online content but no children. DAO links are shown above the tabs; there's no "Online content" tab.
<img width="1017" alt="file - direct content" src="https://user-images.githubusercontent.com/6945691/63987810-59118d00-ca8e-11e9-9596-35a99352065c.png">

gordonleacock:
Some questions about this issue:

1. How can I tell if a component has *direct* "online content" children? Should I use document.digital_objects.present?
2. How can I tell if a component has *indirect* "online content" children? Should I use document.online_content? for this.
3. To display the direct above the tabs, do I just call _<%= render_document_partial(document, 'arclight_viewer') %>_?

gordonleacock:
Status:
1. New issue created related to #680. Show online content for level’s children in a search view in “Online content” tab #927, see #680 for details.
2. Indenting of online content links above the tab is not the same as described in the issue, if this is very important is could be spun out into a separate issue.

PR #922, opened 2019-10-07
Show online content at every level of the hierarchy #680

gordonleacock:
This is a partial solution for ticket #680 
Paired with @labradford and @mejackreed :-)

Yet to do:

- [x] 1. If @jvine would like the direct digital objects above the tab indented that will require some additional code (probably from a blacklight expert).

2. The "Online content" tab should show search results for any children of the current level that have their own direct digital objects and the tab should include a count of those children. But the search and count should not include the direct digital objects.
3. The whole issue will need testing and review by @jvine!

Collection Show Page:
<a href="https://cl.ly/57f207e1bfb1" target="_blank"></a>

Component Show Page:
<a href="https://cl.ly/de28957319f2" target="_blank"></a>

### DIFF
--- a/app/views/catalog/_show_collection.html.erb
+++ b/app/views/catalog/_show_collection.html.erb
@@ -48,7 +48,7 @@
       </div>
       <% if document.online_content? %>
         <div class='tab-pane' id='online-content' role='tabpanel'>
-          <%# TODO: render children with online content in a search result style view %>
+          <%= render partial: 'collection_online_contents', locals: { document: document }  %>
         </div>
       <% end %>
       <div class='tab-pane' id='access' role='tabpanel'>
--- a/app/views/catalog/_show_collection.html.erb
+++ b/app/views/catalog/_show_collection.html.erb
@@ -1,5 +1,14 @@
 <div class='row'>
   <div class='col-md-12'>
+    <% if document.digital_objects.present? %>
+    <p>Test</p>
+        <%= content_tag :p, class: "media" do %>
+            <span class="media-body al-online-content-icon" aria-hidden="true"><%= blacklight_icon :online %></span>
+            <span class="col text-muted"><%= t('arclight.views.show.online_content_upper') %>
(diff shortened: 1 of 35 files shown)


mejackreed:
Since we paired on this, I'm going to ask another team member to review if possible.

app/views/catalog/_show_collection.html.erb
```diff
@@ -38,9 +46,11 @@
       <div class='tab-pane' id='contents' role='tabpanel'>
         <%= render 'collection_contents' %>
       </div>
-      <div class='tab-pane' id='online-content' role='tabpanel'>
-        <%= render partial: 'collection_online_contents', locals: { document: document }  %>
-      </div>
+      <% if document.online_content? %>
+        <div class='tab-pane' id='online-content' role='tabpanel'>
+          <%# TODO: render children with online content in a search result style view %>
```
camillevilla: @gordonleacock Can we retain this partial until someone picks off #927? Removing it now leaves the `Online content` tab empty. It would probably be easier for whoever works on #927 to iterate from that list of links.

labradford: I added these partials back.

labradford:
> @gordonleacock Added functionality here looks fine, but please take a look at my comment re: retaining the Online content partial.
> 
> I also noticed that the breadcrumbs div is starting to do funny things with `margin-bottom`. I would recommend rebasing and incorporating @ggeisler's changes from #928. If the CSS is getting tangled up I'm happy to pair up tomorrow morning.
> 
> breadcrumb-margin

@camillevilla I think the issue with the gap at the bottom of the breadcrumbs is being caused by the context/actions box and not by the margin on the breadcrumbs. 

Maybe we should break this out into a separate ticket? It also occurs when the item has metadata, which needs to appear above the online content items.
<a href="https://cl.ly/785f09d80221" target="_blank"></a>

PR #945, opened 2019-10-11
[WIP] Plus minus integration with collection context

mejackreed:
I'll keep pushing my work here as I get things going.

I've made some progress on dropping in the returned elements into the collapsible div. This work on 1 & 2 level hierarchies but has some bugs at greater hierarchies.

Open to moving this forward more , or ditching in favor of @estelendur approach.

Feel free to drop any question / comments here.


mejackreed:
Closing in favor of #960 #957

PR #951, opened 2019-10-15
Update layout/styles of the search form.

### LINKED ISSUE #722: Responsive / flex adjustments to search bar
Part of #716 

**See comment for current screenshots with scope dropdown as implemented.**

Goal is to give the search input as much room as possible at all sizes. 

- [ ] 1. Add flex layout
- [x] 2. Justify search form
- [x] 3. Remove collapse behaviour on mobile
- [x] 4. Suppress "Search" text at `@media (max-width: 768.98px)`

## 1. Add flex layout
Flex layout controls the placement and size of the scrope dropdown. Here's what worked for me to make the layouts below:

- scope dropdown:  `col-12 col-md-5 col-lg-3`
- search form:  `col-12 col-lg-9`

It's intentional that there's no `col-md-` setting for the search form - it is on a line of its own and fills the full screen width. The `col-md-5` constraint for the scope dropdown prevents it from filling the width, as it doesn't clearly "read" as a control at that size.

**Large: container = 1140; `@media (min-width: 1200px)`**
search 1140

**Large: container = 970 `@media (min-width: 992px)`**
search 960

**Med: container = 720 `@media (min-width: 768px)`**
search 720 final

**Small: container = 540 `@media (min-width: 576px)`**
search 540

**Small: device width = 375**
search 375

## 2. Justify search form
The search form is justified to the full width of its container - that is, the right edge of the Search button should always be flush to the right margin. There's some weirdness around the search box that may make this tricky (or it was for me with in-browser changes) but it looks pretty awkward when it's left jagged.

Screen Shot 2019-09-04 at 4 35 20 PM

## 3. Remove collapse behaviour on mobile

It may make sense to collapse some other header elements (TBD), but search is a core feature, and the relocation of the "search this collection" feature into a collapsed section would hide something that was previously visible. 

## 4. Suppress "Search" text at `@media (max-width: 768.98px)`

Currently the icon-only button starts at `(max-width: 575.98px)`; apply this one level higher, at iPad width.

jkeck:
Closes #722 

## Home Page (before)
home-page-before

## Home Page (after)
home-page-after

---

## Search Results (before)
results-before

## Search Results (after)
results-after

### DIFF
--- a/app/assets/stylesheets/arclight/application.scss
+++ b/app/assets/stylesheets/arclight/application.scss
@@ -1,5 +1,6 @@
 @import 'bootstrap/functions';
 @import 'bootstrap/variables';
+@import 'bootstrap/mixins/breakpoints';
 @import 'bootstrap_overrides';
 @import 'variables';
 @import 'responsive';
--- a/app/assets/stylesheets/arclight/modules/mastheads.scss
+++ b/app/assets/stylesheets/arclight/modules/mastheads.scss
@@ -39,6 +39,27 @@
   }
 }
 
+
+.navbar-search {
+  // Undo some of Blacklight's extending of column classes.
+  .search-query-form {
+    flex-basis: auto !important;
+    max-width: 100% !important;
+  }
+}
+
+.within-collection-dropdown {
+  min-width: 0;
+
+  @include media-breakpoint-up(sm) {
+    width: 50% !important;
+  }
+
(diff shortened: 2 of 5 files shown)


jkeck:
Just an FYI on this, I couldn't get the column layout as specified in the issue, but I think what I've done w/ plain flex/css is to the spirit of what was desired.

Issue #909, opened 2019-10-02
Contents tab should use the same "view type" as collection context


mejackreed:
Currently "Contents" tab uses a different view from collection context. These should be the same.

Contents should still support child expansion and truncation as it did previously.

PR #955, opened 2019-10-16
Add +/- to context view

### LINKED ISSUE #877: Incorporate +/- into new collection context page
See #856 and #824.

estelendur:
Resolves #877.

I'm not sure how much of the messy javascript in `context_navigation.js` can be removed, but I imagine the answer is "at least some of it" and will be investigating that.


estelendur:
Ok, so there's a ton of styling issues:
image

mejackreed:
As a heads up, I'm interested in doing some performance testing on this. My only concern is the page blocking recursive method that relies on Solr.

estelendur:
I made it prettier but I don't know if I messed up existing styling or what other styling we want to do to it, or what. Also it's still screwed up at mobile sizes.

image

mejackreed:
Thanks again for the work on this. We are going to go in a different direction for a few reasons and I wanted to outline them here for our future selves and other ArcLight contributors and also call out some of the things I’ve pulled from this PR out into #957

1. I like how you were able to reuse the +/- markup from before here: https://github.com/projectblacklight/arclight/pull/955/files#diff-c4fbf4eec1570e7dbf366c252deb786dR9-R20. Something similar is implemented here https://github.com/projectblacklight/arclight/pull/957/files#diff-4c44db6c21bc4500997fd3220a1e248bR9-R20
1. While I like the idea of doing as much in Ruby as possible, I still have concerns about using a recursive approach that queries Solr ultimately blocking a page load. Because we can’t predict the complexity that many EAD’s may contain, this approach makes me nervous about unexpected errors in the future which would prevent the page from loading at all. https://github.com/projectblacklight/arclight/pull/955/files#diff-1ad080a75a35bedc7779fbce68d7dcecR294-R305
1. Another complication here, is by directly querying Solr, we are bypassing the chain of Blacklight:SearchService which provides additional functionality (SearchBuilder, doc transformation) expected by Blacklight adopters. Similarly, this would be a new pattern in Blacklight, and while directly referencing the partials works, we also lose the partial lookup chain logic from Blacklight.
1. I’m also not yet clear on the path forward for #909 using this approach and reusing this in other areas of the application that may be needed.

I hate to not move this work forward in its entirety, because it’s really clearly solving the problem at hand, in a way that is likely preferred by many of us (Ruby). I’ve reused several parts of this in #957, but I think with these considerations I don’t see how to make it all fit. Thanks for helping us figure out the boundaries of this problem. This PR helps us better understand one of the options available to us.

PR #832, opened 2019-09-18
Use plus/minus to indicate expansion in document hierarchy

estelendur:
I can't seem to get the image to align with the text/link... Can someone let me know if I need to fuss with the styling more?

Resolves #824.

PR #960, opened 2019-10-18
Fixes #909 by using collection context approach

### LINKED ISSUE #909: Contents tab should use the same "view type" as collection context

mejackreed:
A follow up to #957 (should be reviewed first).

This fixes #909 by enhancing work in #957 and giving us the same semantic structure and view in the Collection Contents tab also.

contents

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -122,6 +122,7 @@ class ContextNavigation {
     this.el = $(el);
     this.data = this.el.data();
     this.parentLi = this.el.parent();
+    this.eadid = this.data.arclight.eadid;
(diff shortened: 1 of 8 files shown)


ggeisler:
@mejackreed This is a very minor issue and I can look at it separately after this is merged if you prefer, but the vertical alignment of the plus and minus icons seems off. I think we solved this in a previous PR; that solution may or may not be relevant here: https://github.com/projectblacklight/arclight/pull/832

mejackreed:
@ggeisler split out to https://github.com/projectblacklight/arclight/issues/961

PR #964, opened 2019-10-18
Remove unused component_contents partial

mejackreed:
I can't figure out how or where this is used.


cbeer:
👍 looks like it got removed in https://github.com/projectblacklight/arclight/pull/912.

PR #972, opened 2019-10-22
Do not show scope label on context nav

mejackreed:
Per @jvine 

## Before
Screen Shot 2019-10-22 at 5 19 45 PM

## After
Screen Shot 2019-10-22 at 5 19 34 PM

### DIFF
--- a/app/views/catalog/_index_collection_context_default.html.erb
+++ b/app/views/catalog/_index_collection_context_default.html.erb
@@ -33,7 +33,11 @@
 
       <div class="order-2">
         <div class="row">
-          <%= render partial: 'arclight_abstract_or_scope', locals: { document: document } %>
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/arclight/modules/context_navigation.scss
```diff
@@ -32,6 +32,7 @@ li.al-collection-context  {
 
   .al-document-abstract-or-scope {
     max-width: 45em;
+    margin-bottom: 0;
```
jkeck: Could/should we use the `mb-0` class on the element here instead?

mejackreed: Oh good idea. I've pushed up this change!

PR #985, opened 2019-12-16
Update icons to use with Blacklight icon update

mejackreed:
See https://github.com/projectblacklight/blacklight/pull/2225


mejackreed:
Replaced by #998

PR #991, opened 2020-03-06
Bugfix: context navigation fetches correct child components when clic…

### LINKED ISSUE #982: Collection Context expand/collapse bug - same components appear as children of different parents
If I am viewing a component page in the demo app ([example](https://arclight-demo.projectblacklight.org/catalog/umich-bhl-2013066aspace_9c83b47189c612469bdf7d4ccb124835)), when I click the plus icon to expand other nearby parent components, I see the same child components listed under every parent (with the current component duplicated and highlighted in yellow multiple times).

Here's a screenshot that better illustrates the problem:

image

seanaery:
…king +/- on a node outside of the originally loaded component's ancestry. Fixes #982.

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -123,20 +123,27 @@ class ContextNavigation {
     this.data = this.el.data();
     this.parentLi = this.el.parent();
     this.eadid = this.data.arclight.eadid;
-    this.originalParents = originalParents || this.data.arclight.originalParents;
+    this.originalParents = originalParents; // let originalParents stay null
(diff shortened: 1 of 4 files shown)


seanaery:
These revisions address a bug in context navigation where from a component view, clicking the [+] expander on a node outside of the current component's ancestry was loading the wrong child nodes into the expanded context navigator.

The main issue was that any ContextNavigation object created via a [+] click was being instantiated with the currently-viewed component's ancestor trail ("originalParents"), which isn't relevant to an arbitrary [+] click. These changes set the originalParents to null for these navigators, and let it stay null so this fallback is reached: using the document ID as the parent in the query that fetches the child nodes to populate the navigator.

### Before
component-context-nav-before

### After
component-context-nav-after

PR #993, opened 2020-04-06
Bugfix: enable collection context expansion for EADs with . in eadid.…

### LINKED ISSUE #984: Collection contents tree is not displaying children for Stanford finding aids
E.g. [Zimbardo](https://arclight-demo.projectblacklight.org/catalog/sc0750-xml): 

Zimbardo

seanaery:
… Fixes #984.

### DIFF
--- a/app/helpers/arclight_helper.rb
+++ b/app/helpers/arclight_helper.rb
@@ -298,7 +298,7 @@ module ArclightHelper
           name: document.collection_name,
           originalDocument: document.id,
           originalParents: original_parents,
-          eadid: document.eadid
+          eadid: normalize_id(document.eadid)
         }
       }
     )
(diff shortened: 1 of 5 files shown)


seanaery:
Note that in order to add a test, I had to add a fixture that has both a dot in the `eadid` and some nested contents to navigate. I grabbed what seemed to be a reasonably small Stanford EAD from the demo app ( https://github.com/sul-dlss/arclight-demo/blob/master/data/ead/sul-spec/pc0170.xml ).

For testing, I also increased the default wait time in the Capybara configuration (from 15 to 25 seconds). I was finding that feature tests like this in the context nav would intermittently fail presumably because the XHR responses with the dynamically loaded content weren't appearing in the DOM in time.

PR #996, opened 2020-04-27
Update normalized_title.rb

archivalGrysbok:
I have many EADs from other institutions that are missing unittitle and unitdate, which causes them to fail indexing entirely, with an unhelpful error message. 
I'd rather have them indexed with "no title" than not index at all.


jkeck:
Hi @archivalGrysbok, thank you fro this contribution.

I believe our current behavior is intentional.  We used to have the ability to supply a default string, but removed that behavior in #435 to address #293.  This change seems to conflict with that.

There is some commentary about this being potentially controversial in that thread, but it seems like this was behavior that was explicitly requested.  I don't know if @mejackreed or @anarchivist have any additional context that they would like to provide.

mejackreed:
I would add the only additional context that I recall is that ArcLight relies on both having an `id` and `unittitle/unitdate` attribute to work correctly. The `id` is configurable I believe. We could look at using the default string approach again or another pattern. 

@archivalGrysbok when you index using this pattern, does the application behave as expected?

anarchivist:
I think there's another few options here:

1) Allow a local indexing override. @jkeck @mejackreed is this an option? I seem to remember there might be a pattern for this but I can't remember how to do it.

2) Implement a `MissingTitleStrategy`, similar to the [`MissingIDStrategy`](https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight#advanced-customizing-behavior-of-indexing-components-wo-ids) we describe here. This feels like a better option to me - and is also in the spirit of 1).

mejackreed:
Option 2 seems like the emerging pattern for something like this in ArcLight, so I would be good with that 👍

archivalGrysbok:
@mejackreed It did behave as expected locally. Here's an [example of an item without a title](http://archives.library.wcsu.edu/caoSearch/catalog/RG005al_802b27a67cfddf4036e1ceb5a04e588186a7fa90). (I've also discovered that the string "no title" is a very common title for objects at some repositories in Connecticut. *sigh*)

I'd also be happy with option 2.

thatbudakguy:
closing since it's resolved by the `MissingTitleStrategy` approach listed above — lmk if this is still an issue.

Issue #812, opened 2019-09-18
Collection context tree: show sibling and parents above current component


camillevilla:
Split out of #670 / #673
Related to #811. 
   - One sibling _above_ the current component is displayed; any siblings above that are hidden, with an Expand button to mark the absence. (2)
   - The immediate parent _above_ the current component is shown; the remainder of the tree to the top is hidden, with an expand button. (3)

![](https://user-images.githubusercontent.com/6945691/63970094-29479280-ca58-11e9-96d0-adbffe79733b.png)

jrgriffiniii:
I'm starting to work to restructure the underlying JavaScript functionality which should enable that the "Expand" buttons are offered for parent/parent-sibling nodes in https://github.com/projectblacklight/arclight/commit/877908f99d121ad9c937607f81d9294f21ba0522

I will likely seek out feedback before progressing too much further with this in order to ensure that these proposed changes are acceptable.  Please note that I am also working directly off of the approach which I proposed in https://github.com/projectblacklight/arclight/pull/889

mejackreed:
Resolved previously in #913.

Issue #673, opened 2019-08-29
Add a "Collection context" tab with interactive Collection navigation


jvine:
Part of #668, also related to #645.

Currently: 
- the Collection show page has a "Contents" tab that displays the collection tree
- other component show pages have a static "Collection context" snapshot on the "Overview" tab

Goal is to:
- create a new tab for "Collection context"
- within that tab, replace the static context snapshot with an interactive tree, open to the current component level

### Before
<img width="500" alt="context before" src="https://user-images.githubusercontent.com/6945691/63968431-88a3a380-ca54-11e9-9170-2c5d0ad3801c.png">

### After
<img width="500" alt="context after" src="https://user-images.githubusercontent.com/6945691/63968432-88a3a380-ca54-11e9-8930-489515df71fb.png">

## Tasks
**Note that all these changes affect non-Collection component pages only.** 
- [x] #810: Add a "Collection context" tab as the first item in the tablist. See #645 for placement 
- [x] Remove the static "Collection context" snapshot
- [x] #810: The Overview tab should now be empty, from changes in #628 and removal of the snapshot, so it can be removed
- [x] **Display the collection tree in the "Collection context" tab**
   - Layout of the tree is as spec'd in #670 
   - Tree is open to show the current component
   - #811: Current component is highlighted with a yellow background (1, below)
   - #811: Because the current component may be far down in the tree, portions of the tree are hidden to ensure the current component is visible in the tab. 
   - #812: One sibling _above_ the current component is displayed; any siblings above that are hidden, with an Expand button to mark the absence. (2)
   - #812: The immediate parent _above_ the current component is shown; the remainder of the tree to the top is hidden, with an expand button. (3)
   - #811:The entire remaining tree _below_ the current component is shown. All siblings of the component (4), siblings of the parent (5), and all remaining series & subseries (6). All components are closed if they are not required to be open to display the current component.
- [x] #813: **Open collapsed sections by clicking Expand** 
   - When the user clicks an Expand button, insert the hidden portion of the tree into that space, and remove the Expand button (there's no parallel Collapse behaviour). Expand (2) inserts all the siblings above the current one. Expand (3) inserts the remainder of the tree above the immediate parent. [Interactive mockup of this behaviour](https://xd.adobe.com/view/fe3abb57-312c-4393-6031-b3da9dba7023-880b/).

[...]

PR #1008, opened 2020-08-12
Bugfix for context nav where a click on an ancestor node's Expand button was omitting its preceding sibling. Fixes #1005.

### LINKED ISSUE #1005: Expand / collapse bug in Collection Context view - some components disappear on expand view
To replicate:

1. View a nested component in a collection that has 2 or more levels of component hierarchy
2. Choose “expand” to view all parent components
3. Notice that the immediate preceding sibling of the context component’s parent does not display. In the example below the Writing Series should appear between the Letters series and the Financial Records series, but is absent.

Screenshot of component view showing missing Writing Series component (parent's immediate preceding sibling) from: https://arclight-demo.projectblacklight.org/catalog/rushbenjaminandjuliaaspace_ca0a788945cca9b7cc56d0e29d1ab554
image

Screenshot of collection view, showing correct list of top-level components
image

CC @seanaery

seanaery:
This bug was caused by the Expand button's click handler not fully accounting for the difference between:
1. The current component level's collapse logic (all preceding siblings are collapsed _except for the preceding one_)
2. The current component's ancestor level's collapse logic (all preceding siblings are collapsed)

...as outlined in #673 & #812 

These changes add a `collapsible` class to any components that _should_ get collapsed/expanded in the current context, then makes the Expand/Collapse buttons (regardless of nesting level) only toggle those.

Before:
arclight-context-before

After:
arclight-context-after

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -67,25 +67,6 @@ class ExpandButton {
 }
 
 /**
- * Modeling <button> Elements which hide or retrieve <li> elements for sibling
- *   documents nested within the <li> elements of the <ul> tree
- * @class
- */
-class NestedExpandButton extends ExpandButton {
-  /**
-   * This retrieves the <li> elements which are hidden/rendered in response to
-   *   clicking the <button> element
-   * @param {jQuery} $li - the <button> element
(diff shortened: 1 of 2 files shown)


jkeck (approved):
Thanks again @seanaery!

PR #1011, opened 2020-08-26
Bugfix for +/- context nav via Microsoft Edge browsers. Display the c…

### LINKED ISSUE #1009: +/- context navigation doesn't work in Microsoft Edge
Reported at Indiana in https://github.com/IUBLibTech/ngao/issues/101 . Clicking the + icon in context navigation fails to load & display the child nodes.

I have observed this as well at Duke using Microsoft Edge 41.x. This appears to be a bug in [`addListenersForPlusMinus()`](https://github.com/projectblacklight/arclight/blob/master/app/assets/javascripts/arclight/context_navigation.js#L359-L360)

seanaery:
…hild nodes and toggle the icon. Fixes #1009.

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -356,8 +356,8 @@ class ContextNavigation {
     const that = this;
     this.ul.find('.al-toggle-view-children').on('click', (e) => {
       e.preventDefault();
-      const targetArea = $($(e.target).attr('href'));
-      if (!targetArea.data().resolved) {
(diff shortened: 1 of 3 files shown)


seanaery:
A couple minor bugs were preventing MS Edge from working. Edge was interpreting `e.target` to be the +/- `svg`; using `e.currentTarget` instead consistently finds the link element whose click event listener fired. The `content` CSS property only works consistently with the `::before` and `::after` pseudo-elements across-browser, so Edge wasn't toggling the +/-. I also removed the hard-coded `+` icon from the view, leaving it up to the CSS rule to render, which it now does consistently.

PR #1019, opened 2021-01-04
Migrate to GitHub actions for CI

jkeck:


jkeck:
Moved commits over to #1020

Issue #1022, opened 2021-01-29
Collection pages fail to load in boilerplate instance with test data


gwiedeman:
A default install of Arclight 0.4.0 fails to show collection pages when using both our data and the example data using Rails 5.2.4.4 and Ruby 2.6.5. The page fails to load and rails server freezes and has to be killed manually. I only have easy access to windows at the moment, but this was the case for both WSL 1 and 2 so I don't *think* its an environment issue. The same result happened for the example EAD and a multiple of our EADs, so I don't suspect its an indexing or data issue.

### To reproduce
1. Use fresh rvm gemset
2. `gem install rails -v 5.2.4.4`
3. `rails new rails new my-app -m https://raw.githubusercontent.com/projectblacklight/arclight/master/template.rb` ([https://github.com/projectblacklight/arclight/wiki/Creating,-installing,-and-running-your-ArcLight-application](https://github.com/projectblacklight/arclight/wiki/Creating,-installing,-and-running-your-ArcLight-application))
4. `cd my-app`
5. `bundle install`
6. `gem install solr_wrapper`
7. `solr_wrapper`
8. (New terminal with same gemset) `mkdir eads` ([https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight](https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight))
9. `wget -P eads/ https://raw.githubusercontent.com/sul-dlss/arclight/master/spec/fixtures/ead/nlm/alphaomegaalpha.xml`
10. `REPOSITORY_ID=nlm traject -u http://127.0.0.1:8983/solr/blacklight-core -i xml -c /home/greg/.rvm/gems/ruby-2.6.5@arctest/bundler/gems/arclight-b1f67c197fad/lib/arclight/traject/ead2_config.rb eads/alphaomegaalpha.xml`
11. `rails s`
12. Browse to [http://localhost:3000/catalog/aoa271](http://localhost:3000/catalog/aoa271)

### Gem List
```
*** LOCAL GEMS ***

actioncable (5.2.4.4)
actionmailer (5.2.4.4)
actionpack (5.2.4.4)
actionview (5.2.4.4)
activejob (5.2.4.4)
activemodel (5.2.4.4)
activerecord (5.2.4.4)
activestorage (5.2.4.4)
activesupport (5.2.4.4)
addressable (2.7.0)
archive-zip (0.12.0)
arel (9.0.0)
[...]

PR #1023, opened 2021-01-29
Replace try with safe navigation

jkeck:
This was also done upstream in projectblacklight/blacklight#2416


jkeck:
This should fix #1022 (based on my local testing).  Our test build does hit the collections page, and the same request was failing in my browser put passing in test.  Not sure what to make of that, but running locally after this change allowed my browser to load collection pages again.  🤷

PR #1026, opened 2021-02-13
Use suggest_index_catalog_path explicitly for the search form's …

### LINKED ISSUE #1025: Autocomplete in Search box on the Repositories page not functioning
When using the search bar on the Repositories index page, the search suggest does not return any results. The suggest function does not work outside of the catalog controller.

See on the Arclight demo app:

https://arclight-demo.projectblacklight.org/repositories

Screen Shot 2021-02-10 at 9 10 55 AM

when it’s working properly, the XHRs are made to e.g.,
https://arclight-demo.projectblacklight.org/catalog/suggest?q=book
but on the repositories page they’re made (incorrectly) to:
https://arclight-demo.projectblacklight.org/catalog?q=book

It looks like the relevant config block is 
https://github.com/projectblacklight/arclight/blob/e39a53bb33127a1b8cc7071c61e000fee4316163/lib/generators/arclight/templates/catalog_controller.rb#L204-L206

Is there a way to get this config into the RepositoriesController?

jkeck:
…autocomplete_path

Closes #1025 

FWIW, this autocomplete will be the autocomplete that spans across the entire index (like it does everywhere else) since filtering autocomplete results based on any current context isn't possible (as far as I understand)

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -21,7 +21,11 @@
       <% end %>
 
       <label for="q" class="sr-only"><%= t('blacklight.search.form.search.label') %></label>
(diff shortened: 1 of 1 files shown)


seanaery (approved):
Looks like a great solution to me -- thanks @jkeck !

labradford (approved):
I have verified this change solves issue #1025 in the Indiana University Arclight project.

PR #1033, opened 2021-08-02
Bugfix for context nav when using Blacklight 7.13.0+ -- configures pa…

### LINKED ISSUE #1032: Component navigation breaks if using Blacklight 7.13.0+
If using a version of Blacklight that is >= `7.13.0`, component navigation will break due to a missing facet field config in `catalog_controller.rb`. See the following for more details/screenshot: https://github.com/sul-dlss/arclight-demo/issues/250 

To remedy the issue, we should add the following to the `catalog_controller.rb` template:
`config.add_facet_field 'parent_ssi', show: false`

Any existing implementations should add that line to their existing `catalog_controller.rb` before updating Blacklight to `7.13.0`+.

seanaery:
…rent_ssi as a hidden facet. Closes #1032.

A followup to https://github.com/projectblacklight/arclight/commit/fd22c860869bcb323bda1efbf9b1d2889ec9f623 -- because `parent_ssi` is used as filter in the URL params for finding child components, it has to be explicitly configured as a facet field for Blacklight versions >= `7.13.0`

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -96,6 +96,9 @@ class CatalogController < ApplicationController
     config.add_facet_field 'places_ssim', label: 'Places', show: false
     config.add_facet_field 'access_subjects_ssim', label: 'Subject', limit: 10
     config.add_facet_field 'component_level_isim', show: false
+
+    # Note that parent_ssim is an array of all ancestor nodes, including the parent
+    # parent_ssi is just the immediate parent; it's used in queries for context nav
     config.add_facet_field 'parent_ssim', show: false
     config.add_facet_field 'parent_ssi', show: false
 
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -97,6 +97,7 @@ class CatalogController < ApplicationController
     config.add_facet_field 'access_subjects_ssim', label: 'Subject', limit: 10
     config.add_facet_field 'component_level_isim', show: false
     config.add_facet_field 'parent_ssim', show: false
+    config.add_facet_field 'parent_ssi', show: false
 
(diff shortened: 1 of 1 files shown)


lib/generators/arclight/templates/catalog_controller.rb
```diff
@@ -97,6 +97,7 @@ class CatalogController < ApplicationController
     config.add_facet_field 'access_subjects_ssim', label: 'Subject', limit: 10
     config.add_facet_field 'component_level_isim', show: false
     config.add_facet_field 'parent_ssim', show: false
+    config.add_facet_field 'parent_ssi', show: false
```
cbeer, line 100: What's the difference between (the content indexed into) `parent_ssim` and `parent_ssi`? Can we consolidate, or at least leave a comment about what they're both used for (or maybe the `ssim` one is a typo)?

seanaery: Thanks @cbeer -- that's a good question, and it's reasonable to assume that could be a typo. `parent_ssim` is an array of the all of the component's ancestor node IDs, from the top-level collection down to its immediate parent. `parent_ssi` is the just final value of the array: the direct parent's ID.

To me, the best way to disambiguate is ultimately to replace the words `parent` or `parents` with `ancestors` throughout the app in all cases except when we truly mean just the immediate parent. That'd mean changes to content/context navigation JS, Solr configs, traject rules, etc. so it'd be a heavier lift that might be better timed with other breaking changes in the future.

I'd be glad to amend this PR with some inline comments to clarify.

cbeer (approved):

lib/generators/arclight/templates/catalog_controller.rb
cbeer: Thanks, that helps my understanding. I've 👍'd this and will happily merge it after giving others (who may have more context for arclight...) a chance to chime in.

PR #1045, opened 2022-10-03
Remove cruft

jcoyne:


jcoyne:
It seems like Rails 6.0.x requires these files.

PR #1046, opened 2022-10-03
Test on blacklight 8

jcoyne:

PR #1057, opened 2022-10-04
Remove Gemfile.extra

jcoyne:
There is no need to provide this

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,5 +41,3 @@ else
   end
 end
 # END ENGINE_CART BLOCK
-
-eval_gemfile File.expand_path('spec/test_app_templates/Gemfile.extra', File.dirname(__FILE__))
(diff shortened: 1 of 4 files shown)


thatbudakguy:
based on slack conversation — was this also blocker for blacklight 8 compatibility? just figured we should note that somewhere

jcoyne:
nope, this just reduces complexity

cbeer:
Is this going to conflict with https://github.com/projectblacklight/arclight/pull/1046?

jcoyne:
@cbeer yes, but I can resolve that by moving it to the Gemfile if necessary.

cbeer:
We can try it. My memory is, it might get tricky pointing at all the right versions of things at the right time, but 🤷‍♂️

PR #684, opened 2019-08-30
633 add access tab

labradford:
Closes issue #633 
Add access tab and move content from sidebar

### Visuals:
Component access tab:
<a href="https://cl.ly/dde73334ac8b" target="_blank"></a>

Collection access tab:
<a href="https://cl.ly/21838c003309" target="_blank"></a>

PR #1074, opened 2022-10-04
Remove unused teplates

jcoyne:


cbeer:
It looks like these were dropped in https://github.com/projectblacklight/arclight/pull/684?

PR #1076, opened 2022-10-04
Move generic_document_fields to the presenter

jcoyne:


cbeer:
I wonder if we can get rid of those methods entirely if we make the presenter just a little smarter? WIP: https://github.com/projectblacklight/arclight/pull/1080

jcoyne:
I think that's where I was heading, albeit in smaller chunks.

PR #1103, opened 2022-10-06
Get rid of the BlacklightFieldConfigurationFactory

cbeer:
I'm not sure I understand what this is doing for us?


app/presenters/arclight/show_presenter.rb
```diff
@@ -43,7 +43,17 @@ def fields
     attr_accessor :field_group
 
     def field_config(field)
-      BlacklightFieldConfigurationFactory.for(config: configuration, field: field, field_group: field_group)
+      return super unless field_group
+
+      fields.fetch(field) do
```
jcoyne, line 48: Won't this now return NullField where before it was returning ShowField?

jcoyne: Oh, nm, that's only if it doesn't find it.

jcoyne (approved):

PR #1119, opened 2022-10-07
Add visually-hidden where sr-only is present

jcoyne:
visually-hidden is the Bootstrap 5 equivalent of sr-only in Bootstrap 4


marlo-longley:
@jcoyne does this imply that we will keep both  BS4 and BS5 syntax in templates? 
Not sure the cleanest path forward for that.

jcoyne:
@marlo-longley yes, this is what is done in blacklight core, I'm just following that pattern.

PR #1121, opened 2022-10-07
Remove unnecessary call to bookmark toggle behavior

jcoyne:
This is not a public method in Blacklight 8


cbeer:
Replaced by #1162

PR #1127, opened 2022-10-10
Reserve document actions for action-like things

cbeer:


app/components/arclight/document_collection_context_component.html.erb
```diff
@@ -34,8 +34,16 @@
...
 
-        <div class="my-w-25 w-md-100 order-12 order-md-1">
-          <%= helpers.render_index_doc_actions document, wrapping_class: 'd-flex justify-content-end' %>
+        <div class="my-w-25 w-md-100 order-12 order-md-1 d-flex justify-content-end">
+          <%= content_tag('div', class: 'al-document-container col col-no-left-padding text-muted ml-auto') do %>
+            <%= document.containers.join(', ') %>
+          <% end if document.containers.present? && !online_contents_context? %>
+
+          <div class="d-inline-flex">
+            <%= tag.div blacklight_icon(:online, classes: 'al-online-content-icon') if document.online_content? %>
```
jcoyne, line 43: FYI, in Blacklight 8 passing classes to `blacklight_icon` will have no effect.  It would be preferable if you just wrapped it in a `<span class="al-online-content-icon">...` as is done here: https://github.com/projectblacklight/arclight/pull/1046/files#diff-b059ac24b59360a52480b21cdbb9fdbcd7d88e1bedf47f2d9dd7be05c74dfeb4R3

jcoyne (approved):

PR #1138, opened 2022-10-11
Require Ruby 3.0+ and Rails 7.0+

jcoyne:


cbeer:
Should we bump the rails version in the gemspec?

jcoyne:
@cbeer yes. I've made the change.

Issue #1148, opened 2022-10-12
Restructure the component-level UI layout


ggeisler:
Since the last community workcycle on ArcLight, several institutions -- [Duke](https://archives.lib.duke.edu/catalog/jwticonographic) and [University at Albany, SUNY](https://archives.albany.edu/description/catalog/ua801-004), for example -- implemented production versions that included a significant change to the UI layout used for display component-level information (as opposed to the search results and home/repositories pages). Other institutions considering implementing ArcLight in the near future have also indicated preferring this updated UI layout.

Below is a very general description of the overall layout change. I think the Stanford developers currently working on ArcLight know how to find the relevant code from Duke to use as a more specific guide, when applicable. I'll also create more specific tickets that describe updates to individual sections that require more description and/or deviate somewhat from Duke's example.

<img width="1023" alt="Screen Shot 2022-10-11 at 5 00 49 PM" src="https://user-images.githubusercontent.com/101482/195220020-0df78027-7b2e-44b8-8254-a16c24d85561.png">

- [ ] Change the component-level show page layout from a single-column to two-columns, where the sidebar column is 1/3rd of the page width, and the main content column is 2/3rds of the page width (i.e., a 4/8 column split) . This is different from the 3/9 column split used on the search results page but 1) we need the sidebar to be wider to accommodate a potentially deeply nested (and therefore indented) collection navigation tree, and 2) I don't think there's any UX problem with having two different sidebar/main content area proportions in these two contexts.

---

More specifically, this comparison shows where the major parts of the UI move from the current version of ArcLight to the recommended updated layout:

<img width="1715" alt="Screen Shot 2022-10-11 at 5 01 33 PM" src="https://user-images.githubusercontent.com/101482/195220072-74691503-30f4-485a-bddc-7ac32f4bab0a.png">

**A: Breadcrumb section** - Moves to the top of the main content area, separated from the component title and details by a bottom border. See #1176 for details.

**B: Collection info** - Moves to the top section of the sidebar column. See #1175 for details.

**C: Collection overview** - This tab and related collection metadata is currently only displayed when you are at the top, collection-level component page. The vertical tabs indicated by the right-side "C" in the current page move to the middle section of the sidebar column, with a section title of `Collection overview`.

[...]

PR #1153, opened 2022-10-12
Extract CollectionComponent and NonCollectionComponent from DocumentC…

### LINKED ISSUE #1150: DocumentComponent has too many conditionals around collection? in the template
Maybe use a different component if it's a collection.

jcoyne:
…omponent

This eliminates a number of conditionals around whether the document is a collection or not.
Fixes #1150


jcoyne:
This may be unnecessary given https://github.com/projectblacklight/arclight/issues/1148

cbeer:
I was hoping to wait on something like this until we were sure how these two differed. It was really hard to see before when it was spread across lots of files, so I imagine some of the differences are just unintentional divergence 🤞

PR #1157, opened 2022-10-12
Use a data attribute for behaviors

jcoyne:
Use the class just for styles. This follows the pattern used by Stimulus for binding elements to controllers.

### DIFF
--- a/app/assets/javascripts/arclight/context_navigation.js
+++ b/app/assets/javascripts/arclight/context_navigation.js
@@ -264,9 +264,9 @@ class ContextNavigation {
     this.el.html(this.ul);
 
     // Initialize additional things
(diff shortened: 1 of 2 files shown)


thatbudakguy (commented):
it seems a little misleading to use stimulus-esque markup when we aren't actually using stimulus for the behavior (yet). could we just wait to make this change until this thing becomes an actual stimulus controller?

cbeer:
I think it's useful to dis-entangle the style (`class`) from the functionality (`data-`), and I think it's useful to adopt their terminology rather than make up our own to mean the same thing🤷‍♂️

thatbudakguy (approved):

PR #1164, opened 2022-10-12
Refactor responsive truncator

### LINKED ISSUE #1091: Replace responsiveTruncate with pure CSS
https://css-tricks.com/line-clampin/

thatbudakguy:
* Removes jQuery dependency
* Shifts truncation logic to SCSS

Closes #1091


thatbudakguy:
there are two notable impacts to using CSS for truncation instead of JS:
- screen reader users don't experience the truncation. to them, the element reads as if it was not truncated at all. thus, we don't show them the toggle button, since it doesn't have any effect for them.
- this gets harder to test, because capybara doesn't know about the truncation of the text (the DOM doesn't change). i'm not sure how to adapt the tests in `spec/features/scope_truncation_spec.rb`.

jcoyne (changes requested):
I believe there is a line in engine.rb that also needs to be removed.

jcoyne (approved):

PR #1180, opened 2022-10-17
Continue primary testing with Blacklight 7

cbeer:


jcoyne:
It looks to me like this suffers from 

https://github.com/projectblacklight/arclight/actions/runs/3267826382/jobs/5373500409#step:5:1143

which was fixed by https://github.com/projectblacklight/blacklight/pull/2834

cbeer:
It doesn't look like Blacklight 7.31 fixed this. Any ideas, or should we revert the 8.x change until we can fix this?

Issue #1174, opened 2022-10-13
Component view layout update: Collection contents


ggeisler:
One of the UI updates noted in https://github.com/projectblacklight/arclight/issues/1148 involves the hierarchical tree for the collection inventory. Currently there is a tab called "Contents" (when at the collection level) or "Collection context" (when at a component level; callout "E" in the screenshots below) that holds this hierarchical tree:

<img width="1870" alt="Screen Shot 2022-10-13 at 2 25 04 PM" src="https://user-images.githubusercontent.com/101482/195713807-86a82209-6736-4df7-bfa1-302285b458b7.png">

These both get replaced by moving the hierarchical tree to the sidebar, under the "Collection contents" heading. The basic function of the tree remains: selecting the expand/collapse icon opens/closes parent levels of the tree, while selecting the title of a component in the tree results in 1) adding a background highlight (I used `#f5f5f5` in the mockup) to indicate it is selected, and 2) loading that component's show page in the main content area.

(Note that if the collection itself is displayed in the main content area, then one of the nav items in the "Collection overview" section should be highlighted. Opening or closing a parent level in the collection contents tree alone shouldn't trigger highlighting of that item; an item should only be highlighted in the tree if it is selected and its show page is displayed in the main content area. In other words, what is showing in the main content area determines which item is highlighted in either the "Collection overview" or the "Collection contents" section.)

For example, here we've used the expand icon to open Series 1 to see its child components, while the collection access info is still displayed in the main content area:

<img width="1201" alt="Screen Shot 2022-10-13 at 2 33 13 PM" src="https://user-images.githubusercontent.com/101482/195715224-bdcfca30-ba2a-4cda-9a00-798f7a5af341.png">

---

And now we've selected the "Series 1" title itself, which loads the Series 1 component in the main content area:

<img width="1012" alt="Screen Shot 2022-10-13 at 1 52 42 PM" src="https://user-images.githubusercontent.com/101482/195711679-9dfd9de5-8d5a-4786-bcb6-145add111643.png">

---

Here we've selected one of the child components of Series 1 in the tree, which loads its show page in the main content area: 

<img width="1011" alt="Screen Shot 2022-10-13 at 2 01 49 PM" src="https://user-images.githubusercontent.com/101482/195711880-e6db3cb6-9c17-450a-9339-ee70b2ef0fec.png">

---

[...]

PR #1191, opened 2022-10-18
Remove the description from the collection hierarchy display

cbeer:


jcoyne:
@cbeer can you say why you made this change?

cbeer:
This is part of  Gary's designs in #1174 and looks particularly bad right now?

PR #1193, opened 2022-10-18
Remove javascript that was for the online tab count

jcoyne:
This element was removed in #1173, so this javascript became inoperative


cbeer:
I don't think this is gone -- the oembed viewer is provided by Arclight::EmbedComponent

jcoyne:
@cbeer I didn't see this data attribute provided by that: https://github.com/projectblacklight/arclight/pull/1193/files#diff-4029212770df71c01c0bf678b6b20de2313975a0c1bf4f1bee2206f1392c73bbL4

cbeer:
Ah, I must have missed that. The content of the online tab moved elsewhere. I think something like https://github.com/projectblacklight/arclight/pull/1199 is the appropriate fix here.

jcoyne:
@cbeer I merged yours, then rebased mine. This still has code about the online count that ought to be removed.

PR #1203, opened 2022-10-19
Redo breadcrumbs to match updated designs

### LINKED ISSUE #1176: Component view layout update: Breadcrumbs
One of the UI updates noted in https://github.com/projectblacklight/arclight/issues/1148 involves the breadcrumb display at the top of the component page. Currently we display each level of the breadcrumb for the current item on a new line, indenting slightly for each line, and prepending each line with an icon representing the type item it is.

Because there can be so many levels of nesting, this approach -- while making each level of the breadcrumb very visually distinct -- can use up a lot of vertical space. For example:

Screen Shot 2022-10-14 at 2 39 04 PM

Duke's ArcLight implementation modified this approach. They place the collection title on its own line (their implementation does not display the owning repository, as ArcLight does by default), with each successive level separated by an angle bracket icon on the same line, unless the item title cannot completely fit on the remaining space on that line, in which case it breaks to a new line, beginning with the angle bracket icon. They have also dispensed with the item icons, aside from the collection icon. For example, this item in Duke's collection is eight levels deep, similar to the ArcLight fixture example above, but takes up much less vertical space:

Screen Shot 2022-10-14 at 2 40 01 PM

This approach seems worth updating ArcLight to, given the savings in vertical space with arguably only a slight reduction in ease of visually parsing.

Because ArcLight needs to include the repository in the breadcrumb (we know many institutions do want to indicate the owning repository), the recommendations below vary a bit from the Duke example, but the general approach is very much "do what Duke did". This is certainly open for discussion, though (I admit I don't completely love this solution; including the repository makes it more difficult to keep the clean look of Duke's).

--- 

Screen Shot 2022-10-14 at 3 18 03 PM

---

- [ ] **A**: Continue to display the repository and collection parts of the breadcrumb as we do now, using the same icons and font sizes
- [ ] **B**: At the first level below the collection, use an arrow icon (see end of ticket) to indicate we are starting to show items from the collection, followed by the first level item title. Alignment might be tricky but ideally the edges of the collection and first level titles are left-aligned.
- [ ] **C**: Each subsequent level is represented by an right angle bracket icon followed by the item title. If those elements fit on the current line without wrapping, they go on that line, as in the line with the callout "D". But if they don't, they should break to a new line, as in the lines with the callouts "C" and "E".
- [ ] In my mockup I separated the lines of the breadcrumb by `0.5rem`, so that might be starting point to try.

## Icons
- Right arrow icon: see [IU's ArcLight implementation](https://archives.iu.edu/catalog/VAA4026aspace_VAA4026-0003) where they use it in their otherwise core ArcLight breadcrumb
```
<svg width="16" height="16" fill="currentColor" class="bi bi-arrow-return-right" viewBox="0 0 16 16" aria-label="Arrow Return Right" role="img"><title>Arrow Return Right</title>
  <path fill-rule="evenodd" d="M1.5 1.5A.5.5 0 0 0 1 2v4.8a2.5 2.5 0 0 0 2.5 2.5h9.793l-3.347 3.346a.5.5 0 0 0 .708.708l4.2-4.2a.5.5 0 0 0 0-.708l-4-4a.5.5 0 0 0-.708.708L13.293 8.3H3.5A1.5 1.5 0 0 1 2 6.8V2a.5.5 0 0 0-.5-.5z"></path>
</svg>
```
- Angle bracket icon: https://fontawesome.com/icons/angles-right?s=solid&f=classic

## Examples from ArcLight fixtures

Screen Shot 2022-10-14 at 3 13 36 PM

---

Screen Shot 2022-10-14 at 3 18 58 PM

---

Screen Shot 2022-10-14 at 3 19 19 PM

---

Screen Shot 2022-10-14 at 4 00 41 PM

jcoyne:
Fixes #1176

Screen Shot 2022-10-19 at 12 35 10 PM

### DIFF
--- a/lib/generators/arclight/templates/arclight.scss
+++ b/lib/generators/arclight/templates/arclight.scss
@@ -1,2 +1,3 @@
+$breadcrumb-divider: quote("»");
 @import 'bootstrap';
 @import 'arclight/application'
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -4,19 +4,33 @@
   margin-bottom: $spacer;
 }
 
-.al-show-breadcrumb .breadcrumb {
-  background-color: transparent;
-  margin-bottom: 0;
-  padding: 0;
+.al-show-breadcrumb {
+  border-bottom: $border-width solid $border-color;
+  padding-bottom: $spacer * .5;
+  margin-bottom: $spacer * .75;
 
-  .breadcrumb-item + .breadcrumb-item {
-    padding-top: 2px;
+  .breadcrumb {
+    .breadcrumb-item .breadcrumb {
+      padding-left: 0.5rem;
+    }
 
-    &::before {
-      content: '';
+    .breadcrumb-item-1 > .breadcrumb {
+      padding-left: 1rem;
+    }
+
+    .breadcrumb-item-3:first-of-type {
+      &::before {
+        content: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='20' height='15'><path fill-rule='evenodd' d='M1.5 1.5A.5.5 0 0 0 1 2v4.8a2.5 2.5 0 0 0 2.5 2.5h9.793l-3.347 3.346a.5.5 0 0 0 .708.708l4.2-4.2a.5.5 0 0 0 0-.708l-4-4a.5.5 0 0 0-.708.708L13.293 8.3H3.5A1.5 1.5 0 0 1 2 6.8V2a.5.5 0 0 0-.5-.5z'></path></svg>"));
+      }
     }
   }
+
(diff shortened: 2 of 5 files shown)


lib/generators/arclight/templates/arclight.scss
```diff
@@ -1,2 +1,3 @@
+$breadcrumb-divider: quote("»");
```
corylown, line 1: When trying this out locally with a freshly generated .internal-test-app it seems that this variable is not getting applied when set in `arclight.scss` and the default Bootstrap `/` breadcrumb divider is displaying instead. It does work if I add `$breadcrumb-divider: quote("»");` at the top of `blacklight.scss`. Maybe there's something wrong with my local setup?

jcoyne: Yes, I think so. `blacklight.scss` should be deleted via https://github.com/projectblacklight/arclight/pull/1202/files#diff-68e5bfaa20d28433265c9d3d44f4065954fde0d86389bbf2dce3a2ea61fa0730R46

PR #1211, opened 2022-10-20
Remove JQuery based element from contextNavigation

jcoyne:


app/assets/javascripts/arclight/context_navigation.js
```diff
@@ -101,9 +101,9 @@ class Placeholder {
 
 class ContextNavigation {
   constructor(el, originalParents = null, originalDocument = null) {
-    this.el = $(el);
-    this.data = this.el.data();
-    this.parentLi = this.el.parent();
+    this.el = el
+    this.data = $(this.el).data();
```
thatbudakguy, line 105: can we not do `this.el.dataset` here?

jcoyne: That's already done in #1210

thatbudakguy: got it. I think if this gets rebased and conflicts are resolved it'll be a little easier to review

jcoyne: It has been rebased.

thatbudakguy (approved):

PR #1216, opened 2022-10-20
Remove jquery from collection navigation

jcoyne:


app/assets/javascripts/arclight/collection_navigation.js
```diff
@@ -60,21 +60,23 @@
...
+      CollectionNavigation.init(element)
+    })
 
-    $('[data-controller="arclight-contents"]').on('navigation.contains.elements', function (e) {
-      $(e.target).find('.collapse').on('show.bs.collapse', function (ee) {
-        var $newTarget = $(ee.target);
-        $newTarget.find('[data-controller="arclight-contents"]').each(function (i, element) {
-          CollectionNavigation.init(element); // eslint-disable-line no-undef
-        // Turn off additional ajax requests on show
-          $newTarget.off('show.bs.collapse');
```
thatbudakguy, line 76: just for my own understanding — does `{ once: true }` achieve the same effect as using jQuery's `.off()` here?

jcoyne: I believe it does. We have set up a listener that removes itself after the first time that it's called.

thatbudakguy (approved):

PR #1218, opened 2022-10-20
Use subqueries to retrieve collection-level data

cbeer:


app/models/concerns/arclight/solr_document.rb
```diff
@@ -34,12 +28,8 @@ def parent_levels
...
     end
 
-    def parent_document
-      self.class.new fetch('parent').fetch('docs', []).first
-    end
-
     def eadid
-      fetch('ead_ssi', nil)&.strip
+      first('ead_ssi')&.strip || collection&.first('ead_ssi')&.strip
     end
```
jcoyne, line 33: can you add a comment about why you fall back to looking at the collection for ead_id?

app/models/concerns/arclight/solr_document.rb
```diff
@@ -51,19 +41,25 @@ def unitid
...
 
     def repository_and_unitid
       [repository, unitid].compact.join(': ')
     end
 
+    def collection
+      return self if collection?
+
+      @collection ||= self.class.new(self['collection']&.dig('docs', 0), @response)
```
jcoyne, line 54: can you add a comment about what this is doing?

app/models/concerns/arclight/solr_document.rb
```diff
@@ -107,7 +103,7 @@ def level
     end
 
     def collection?
-      level.parameterize == 'collection'
+      level&.parameterize == 'collection'
```
jcoyne, line 106: why wouldn't this have a level?

app/models/concerns/arclight/solr_document.rb
```diff
@@ -51,19 +41,25 @@ def unitid
...
 
+    def collection
+      return self if collection?
+
+      @collection ||= self.class.new(self['collection']&.dig('docs', 0), @response)
+    end
+
     def collection_name
-      first('collection_ssm')
+      collection&.normalized_title
```
jcoyne, line 58: is there a reason it wouldn't have a collection?

jcoyne (changes requested):

app/models/concerns/arclight/solr_document.rb
cbeer: It's probably bad test data, but it's not impossible 🤷‍♂️

app/models/concerns/arclight/solr_document.rb
cbeer: Most likely, bad fixture data. There's some indication in the indexing code that the top-level EAD might be something other than a collection, though, so it didn't feel too inappropriate to me 🤷‍♂️

jcoyne (approved):

Issue #1175, opened 2022-10-13
Component view layout update: Collection info


ggeisler:
One of the UI updates noted in https://github.com/projectblacklight/arclight/issues/1148 involves the collection information box, which is currently on the top right of any collection or component page: 

<img width="975" alt="Screen Shot 2022-10-13 at 4 09 09 PM" src="https://user-images.githubusercontent.com/101482/195726870-3c2c0ef0-c532-424e-aa23-4996ea43a37b.png">

In the revised layout, the proposal is to move this box to the top of the sidebar column, with some changes to the components of the box:

<img width="1198" alt="Screen Shot 2022-10-13 at 4 12 48 PM" src="https://user-images.githubusercontent.com/101482/195727301-66897e82-1d94-49f2-9eb7-150bff19494c.png">

## Box styling recommendations
We can start with the current box color and border but adjust the text:
- `Collection` heading: `h3` at `22px` (`1.375rem`), medium weight
- Title of the collection: regular paragraph text at `1rem` medium weight

## Collection ID
Remove this because it is now displayed as part of the collection admin metadata dropdown, described at the end of this ticket.

## Bookmark checkbox
Remove this because it is now displayed next to the collection/component title.

## Request button
- Size: `btn-sm`
- Color: `primary` (the presence and use of this button is likely to vary a lot between institutions, so this is basically just a placeholder)

## Download button
This is a significant change from the existing download links. The issue here is I think we want to keep the collection info box as compact as possible so it doesn't take unnecessary vertical space away from the other, likely more frequently-used sidebar sections. The challenge is, what is downloadable for a given collection is highly variable, depending on the institution policies and what might have been created for a given collection. In our last ArcLight workcycle it seemed that most collections had either a PDF of the finding aid, an XML of the EAD, both, or neither. So the proposal (which is certainly open to other ideas) here is:

If there are no downloads for a collection, don't show a download button:

<img width="397" alt="Screen Shot 2022-10-13 at 4 33 46 PM" src="https://user-images.githubusercontent.com/101482/195730030-f6d75118-cc20-4f83-aed2-ad1c0be1ae12.png">

If there is more than one download available for the collection, show a "Download" dropdown, with each download option in the dropdown menu (ideally using the icons we currently use for the download links; the icons in the mockup are just placeholders for the PDF/XML icons):

[...]

PR #1221, opened 2022-10-21
Display the collection info dropdown

jcoyne:
Ref #1175
Screen Shot 2022-10-21 at 1 44 48 PM

### DIFF
--- a/app/components/arclight/collection_info_component.rb
+++ b/app/components/arclight/collection_info_component.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Arclight
+  # Render information about the collection
+  class CollectionInfoComponent < ViewComponent::Base
+    def initialize(collection:)
+      super
+
+      @collection = collection
+    end
+
+    attr_reader :collection
+
+    delegate :total_component_count, :online_item_count, :last_indexed, :collection_unitid, to: :collection
+
+    def info_icon
+      icon = <<~SVG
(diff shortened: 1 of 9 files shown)


app/components/arclight/collection_info_component.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      @collection = collection
+    end
+
+    attr_reader :collection
+
+    delegate :total_component_count, :last_indexed, :collection_unitid, to: :collection
+
+    def online_item_count
+      Arclight::OnlineItemCount.for(collection: collection)
```
cbeer, line 17: Can we avoid the extra solr query and either index this too, or get Solr to calculate it as part of the collections subquery?

jcoyne: Yes, but I'm not sure how one would do that.

cbeer: I think the existing `has_online_content_ssim` probably isn't what we want, now that I look at it -- it applies to any element in the hierarchy with a `dao` node, not just the component that actually has a DAO.

I wonder if this is supposed to count components (e.g. the c + c1...c12 nodes that contain a `did/dao`) or the actual DAOs.

cbeer: From Gary:

> My thought is that “Total components” is the count of component objects, not factoring in how many digital objects they might have, and “Items available online” is the total count of digital objects in the collection, not factoring in how many components there are.

So I think we have to index a new value at the collection level (the count of `//dao` elements?), and can just use that.

PR #1248, opened 2022-10-31
Simplify Embed markup and align with the Access section

jcoyne:


app/components/arclight/embed_component.html.erb
```diff
@@ -1,13 +1,10 @@
-<%= tag.p class: "media" do %>
-  <span class="media-body al-online-content-icon" aria-hidden="true"><%= helpers.blacklight_icon :online %></span>
-  <span class="col text-muted text-uppercase">
-    <%= t('arclight.views.show.embedded_content') %>
-  </span>
+<div id="embed">
+  <h2 class="al-show-sub-heading"><%= t('arclight.views.show.embedded_content') %></h2>
```
thatbudakguy, line 2: do we for sure want to get rid of the "online content" icon?

app/components/arclight/document_component.html.erb
```diff
@@ -31,12 +31,7 @@
 </div>
 
 <%= render Arclight::MetadataSectionComponent.with_collection(component_metadata_partials, metadata_attr: { layout: Arclight::UpperMetadataLayoutComponent }, presenter: presenter, classes: ['al-metadata-section']) unless document.collection? %>
-
-<div class='row'>
-  <div class='col-md-12'>
-    <%= embed %>
-  </div>
-</div>
+<%= embed %>
```
thatbudakguy, line 34: i think this needs 1rem of margin to separate it from the metadata section below

app/components/arclight/embed_component.html.erb
jcoyne: It's not in Gary's mockups, so my thought was that he didn't want it.

app/components/arclight/document_component.html.erb
jcoyne: I added this in the EmbedComponent

app/components/arclight/embed_component.html.erb
thatbudakguy: i agree, but it seems different to make that decision at the level of arclight vs. just hiding or overriding it in NTA

ggeisler: Yes, @jcoyne is correct that I didn't want the icon, but @thatbudakguy is correct that we probably shouldn't make that decision for everyone (my mockup was in an NTA ticket so it didn't occur to me to clarify that).

jcoyne: Since this is for a new major release (1.0.0) and Gary is the designer of the original and because we're making many other visual changes, it seems like now is the ideal time to push in a change like this.

jcoyne: Since this is for a new major release (1.0.0) and Gary is the designer of the original and because we're making many other visual changes, it seems like now is the ideal time to push in a change like this.

cbeer: It should be easy enough for downstream implementations to re-add an icon (e.g. in css).

PR #1253, opened 2022-11-01
Provide a docker-compose that can run solr

jcoyne:
This approach has the advantage that you don't need to re-index every time you restart the solr service.


marlo-longley:
@jcoyne this works for me and seems useful, I just want to say that you can persist data on shutdown with `solr_wrapper` using the `--persist` flag.

marlo-longley (approved):

PR #1254, opened 2022-11-01
Fix icons for series/collection breadcrumbs

### LINKED ISSUE #1238: Breadcrumb icons show up one step late
It looks like there is a slight glitch with the way the revised breadcrumb display works. The icon associated with each of the top-level breadcrumb nodes appears associated with the linked version of that node, rather than with the node itself. In other words, the icon associated with the node does not appear until we are looking at the child component of that node.

For example, below we are at the top-level collection component. Note that no icon is displayed next to its breadcrumb node:

Screen Shot 2022-10-27 at 9 45 43 AM

---

When we go one level deeper to a series, a direct child of the collection, the icon for the collection breadcrumb node does appear, but the series node does not display an icon:

Screen Shot 2022-10-27 at 9 45 58 AM

---

And then when we go one more level deep to a child of the series, the series icon appears:

Screen Shot 2022-10-27 at 9 46 21 AM

---

Desired behavior is that the icons for the breadcrumb nodes representing the collection and the first child below the collection display when we are on those respective component show pages, not waiting until we are on their child's component page.

jcoyne:
when on the page for a series or collection page.

Fixes #1238
Screenshot 2022-11-01 at 10 15 59 AM
Screenshot 2022-11-01 at 10 15 50 AM
Screenshot 2022-11-01 at 10 15 43 AM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -9,6 +9,8 @@
   padding-bottom: $spacer * .5;
   margin-bottom: $spacer * .75;
   width: 100%;
+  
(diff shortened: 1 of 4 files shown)


ggeisler:
@jcoyne That looks good, except I'm wondering why in your examples the Repository node doesn't appear to be a link? 

Maybe this is just an artifact of your dev environment or something (or an issue unrelated to the goal of this PR) and it actually is a link, but it caught my eye because I believe every node except for the current page should be a link.

jcoyne:
@ggeisler that's just because the indexed solr documents don't have a matching repository in config/repositories.yml

ggeisler:
Okay, 👍

corylown (approved):

PR #1256, opened 2022-11-01
Configure index task to use rails environment

### LINKED ISSUE #1255: Indexing rake tasks don't use Blacklight's default Solr connection

marlo-longley:
Fixes #1255


cbeer:
It looks like the gem-level rake tasks try to run `arclight:index` within the context of the gem:

https://github.com/projectblacklight/arclight/blob/main/tasks/arclight.rake#L60

I'm not sure if that's a feature, or if we can just wrap it with `within_test_app` (and adjust the paths accordingly) so we're running it in the app's context?

jcoyne (approved):

Issue #1223, opened 2022-10-21
Add release notes for v1


marlo-longley:

jcoyne:
This should include:

https://github.com/projectblacklight/arclight/blob/aabfad7b7d9b2a568964dc0f50112c1b454bacdf/lib/generators/arclight/install_generator.rb#L80-L91

marlo-longley:
Ability to configure date normalizer for indexing (#1344)

PR #1257, opened 2022-11-01
Show the collection downloads in the collection context

jcoyne:
Previously this was showing downloads for whatever the current unitid was. 

Also renamed DocumentActionsComponent to CollectionContextComponent


marlo-longley:
@jcoyne this seems like a good candidate to add to https://github.com/projectblacklight/arclight/issues/1223 - v1 release notes?. If I understand from discussion correctly it will undo work that was explicitly added here: https://github.com/projectblacklight/arclight/commit/1e2295583762ca6fc9755aa7ce475bac9a84580e

corylown (approved):

PR #1262, opened 2022-11-02
Adds :hierarchy and :nest_path to search_state_fields configuration

corylown:
Resolves errors such as:

```
Unpermitted parameters: :hierarchy, :nest_path. Context: { controller: CatalogController, action: hierarchy, request: #<ActionDispatch::Request:0x000000011497de10>, params: {"hierarchy"=>"true", "nest_path"=>"/components#5/components#0/components#3", "controller"=>"catalog", "action"=>"hierarchy", "id"=>"mt839rq8746"} }
```

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -236,7 +236,7 @@ class CatalogController < ApplicationController
     end
 
     # These are the parameters passed through in search_state.params_for_search
-    config.search_state_fields += %i[id group hierarchy_context original_document]
+    config.search_state_fields += %i[group hierarchy hierarchy_context id nest_path original_document]
     config.search_state_fields << { original_parents: [] }
 
     # "sort results by" select (pulldown)


lib/generators/arclight/templates/catalog_controller.rb
```diff
@@ -236,7 +236,7 @@ class CatalogController < ApplicationController
     end
 
     # These are the parameters passed through in search_state.params_for_search
-    config.search_state_fields += %i[id group hierarchy_context original_document]
+    config.search_state_fields += %i[group hierarchy hierarchy_context id nest_path original_document]
```
jcoyne, line 239: Do we know why nest_path has to be in the search state?

jcoyne: Looks like it's used to see if the document is in the current hierarchy: https://github.com/projectblacklight/arclight/blob/main/app/components/arclight/document_collection_context_component.rb#L42-L48

jcoyne (approved):

lib/generators/arclight/templates/catalog_controller.rb
cbeer: I don't think they need to be in search state if we're just plucking them out of params directly, but I guess there's not a way to let Rails know we don't care 🤷‍♂️

jcoyne: Right, so these have nothing to do with the "search state", they are only used for display options.

jcoyne: @cbeer are you thinking something like https://github.com/projectblacklight/arclight/pull/1265 instead?

jcoyne:
I think this is resolved by #1265

PR #1265, opened 2022-11-02
Create the search state without display related parameters

jcoyne:


corylown (approved):
This seems like a good approach.

PR #1268, opened 2022-11-03
Add seam for specifying our own component indexing config

cbeer:


jcoyne (commented):
I'm pretty sure this fix is needed in order to make this pass: https://github.com/projectblacklight/blacklight/pull/2868

jcoyne (approved):

PR #1274, opened 2022-11-04
Use offcanvas for the sidebar for small screens

cbeer:


thatbudakguy:
I ended up not being able to put the request buttons in between title and bookmark anymore; in order to get the right DOM order to have them move _below_ the title the bookmark would need to be last, but if that happens it can float above and cover the title text on certain screens.

I went with the simplest approach: just put all the buttons together below the title. I think the most common case is to have items be requestable only one way, so we'll usually see a primary "request" button and a secondary "open the sidebar" button together (the first pic below). If we want the sidebar button on its own line instead, that's not hard to do — it just seemed like a lot of whitespace, even on mobile.

# pictures
a typical case where an item is requestable

Screen Shot 2022-11-08 at 11 46 59

an unusual case where an item has two request options; showing there's still space even on the smallest screen
Screen Shot 2022-11-08 at 11 45 52

the sidebar as it normally appears on large screens
Screen Shot 2022-11-08 at 12 06 27

the sidebar as it appears when summoned from off-canvas on small screens, scrollable, with close button
Screen Shot 2022-11-08 at 12 06 39

cbeer:
@ggeisler : Is moving the request button to the next line ok, or maybe this is a good candidate for css grid?

thatbudakguy:
OK, here's the grid solution I came up with. On large & up, the request button is with the title/bookmark and the sidebar toggle is hidden. Below large, the request buttons are on their own line, and the sidebar toggle is below that. Shown with firefox grid inspector for clarity.

large
Screen Shot 2022-11-09 at 13 48 20

medium/small
Screen Shot 2022-11-09 at 13 48 29

xs
Screen Shot 2022-11-09 at 13 48 50

app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
```diff
@@ -103,6 +103,38 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
...
+  grid-template-areas:
+    "title bookmark-toggle"
+    "request request"
+    "sidebar-toggle sidebar-toggle";
+  justify-content: start;
+  align-items: start;
+
+  @include media-breakpoint-up(lg) {
+    grid-template-areas:
+      "title request bookmark-toggle"
```
cbeer, line 117: Should we stick `sidebar-toggle` in here any way? I guess I'm worried about downstream apps that might hide want to always hide that sidebar?

```suggestion
      "title request bookmark-toggle"
      "sidebar-toggle sidebar-toggle sidebar-toggle"
```

thatbudakguy: Sure, no harm in doing that

corylown (approved):

PR #1276, opened 2022-11-04
Style bookmarks on the search results page

jcoyne:
The same way they are styled on the show page.

Before:
Screenshot 2022-11-04 at 4 49 52 PM

After:
Screenshot 2022-11-04 at 4 50 33 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -103,23 +103,20 @@
   }
 }
 
-.bookmark-toggle .toggle-bookmark {
-  display: inline;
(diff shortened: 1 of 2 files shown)


ggeisler:
I don't like saying this after you've done the work, but I'm not sure we want to do this? Was it requested by someone?

I think the show page styling around the bookmarks is good, because there's room available and it highlights their availability a bit. 

But for search results, I don't think that highlighting is necessary and it increases the height of the result item elements a bit, which isn't ideal.

jcoyne:
@ggeisler no problem. I thought the inconsistency was  a bug not a feature.

PR #1277, opened 2022-11-05
Extract search result breadcrumbs to a component

jcoyne:
So that individual implmentations can customize this component


app/components/arclight/search_result_breadcrumbs_component.html.erb
```diff
@@ -0,0 +1,7 @@
+
+        <% rendered_breadcrumbs = capture { render breadcrumbs } %>
+        <% if rendered_breadcrumbs.present? %>
```
cbeer, line 3: Can we pull this conditional up to the `Arclight::SearchResultBreadcrumbsComponent#render?` `Arclight::BreadcrumbComponent` might need one too.

PR #1279, opened 2022-11-07
Make Arclight::SolrDocument a concern

jcoyne:
This matches the upstream pattern (https://github.com/projectblacklight/blacklight/blob/main/app/models/concerns/blacklight/document.rb\#L19) and allows us to use class_attributes upstream like in https://github.com/projectblacklight/blacklight/pull/2872/files


cbeer:
Hm. It looks like this change makes it impossible to use the attribute in a downstream override:

```
super: no superclass method `extent' for #<SolrDocument:0x000000010cb61540 
```

PR #1283, opened 2022-11-07
Read traject config file from CONFIG_PATH

marlo-longley:


jcoyne:
At this point does it still make sense to use this rake task?  Why not just call traject by itself?

marlo-longley:
@thatbudakguy and I were wondering that initially too. But this seemed like a simple change and one others might benefit from too especially now that it's easier to add custom config files with that seam.

jcoyne:
@marlo-longley I guess I don't see why you'd want to use this at all. It's more indirection (read slower) and poorer error handling.   I'm not sure what the benefit is.

marlo-longley:
@jcoyne I guess I meant the minor benefit of others being able to set CONFIG_PATH and still use Arclight's rake tasks.

Are you saying you prefer to rewrite new indexing tasks for VT/NTA?

cbeer:
I'd tend to agree with @jcoyne; passing mysterious environment variables isn't a great pattern for extensibility. I'm not seeing  so much going on here that calling traject directly or through an application-specific rake task creates much of a burden. 

Alternatively, maybe we could store a path as part of the `Arclight::Engine.config` and let downstream applications change that configuration?

PR #1284, opened 2022-11-07
Fix positioning of document tools when the title is short

### LINKED ISSUE #1282: Search results bookmark alignment can be off
It looks like something is off with the way we align the search result item elements in the title row, such that if an item has a relatively short title, the bookmark element isn't right-aligned the way items with longer titles are:

Screen Shot 2022-11-07 at 3 10 26 PM

You can see this by selecting the "Collections" main menu item to get a collections result set (in either the All results or Grouped by collection views).

jcoyne:
Fixes #1282


app/components/arclight/search_result_component.html.erb
```diff
@@ -43,5 +38,9 @@
         <% end %>
       </div>
     </div>
-  </div>
+
+
+    <div class="col-sm-2">
+      <%= helpers.render_index_doc_actions document, wrapping_class: 'd-inline-flex justify-content-end' %>
```
corylown, line 44: If the intent is for the document action controls to align right, I think the bootstrap flex/content alignment classes should be on the column div? I'm not sure if that's what's intended or not though.

As written:
Screen Shot 2022-11-08 at 9 41 12 AM

My proposed change:
Screen Shot 2022-11-08 at 9 41 43 AM

```suggestion
    <div class="col-sm-2 d-inline-flex justify-content-end">
      <%= helpers.render_index_doc_actions document %>
```

jcoyne: No, the intention is not for them to align right, but to align left to the grid at +10.

corylown (approved):

PR #1287, opened 2022-11-08
Remove unused code

jcoyne:


cbeer:
Wait, what are we using instead?

jcoyne:
My mistake, this isn't unused, it's overriding Blacklight.

PR #1288, opened 2022-11-08
Align the Arclight componets with the blacklight components

jcoyne:
This makes the search results match up more directly with the Blacklight components so that it is easier to understand how the two gems relate to each other.

### DIFF
--- a/app/components/arclight/search_result_component.html.erb
+++ b/app/components/arclight/search_result_component.html.erb
@@ -7,33 +7,14 @@
   itemscope: true,
   itemtype: @document.itemtype,
   class: classes.flatten.join(' ') do %>
-    <div class="col-auto">
+    <div class="px-2">
       <%= icon %>
     </div>
+    <div class="container">
+      <%= render Arclight::SearchResultTitleComponent.new(document: document, compact: compact?) %>
 
-    <div class="col-sm-9 col-no-left-padding d-flex flex-wrap">
-      <div class="documentHeader my-w-75 w-md-100 order-0" data-document-id="<%= document.id %>">
-        <h3 class="index_title document-title-heading">
-          <%= helpers.link_to_document document, counter: @counter %>
-          <%= content_tag('span', class: 'al-document-extent badge') do %>
-            <%= document.extent %>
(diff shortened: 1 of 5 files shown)


app/components/arclight/search_result_component.html.erb
```diff
@@ -7,33 +7,14 @@
   itemscope: true,
   itemtype: @document.itemtype,
   class: classes.flatten.join(' ') do %>
-    <div class="col-auto">
+    <div class="px-2">
       <%= icon %>
     </div>
```
corylown, line 12: Maybe icon should move to the `SearchResultTitleComponent` so it can stay inline with the title?
```suggestion
```

app/components/arclight/search_result_title_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<header class="documentHeader row" data-document-id="<%= @document.id %>">
+  <h3 class="index_title document-title-heading col">
+    <%= helpers.link_to_document @document, counter: @counter %>
```
corylown, line 3:
```suggestion
    <span class="px-2">
      <%= icon %>
    </span>
    <%= helpers.link_to_document @document, counter: @counter %>
```

app/components/arclight/search_result_title_component.rb
```diff
@@ -0,0 +1,15 @@
...
+  class SearchResultTitleComponent < Blacklight::DocumentTitleComponent
+    def initialize(compact:, **args)
+      @compact = compact
+      super(**args)
+    end
+
+    def compact?
+      @compact
+    end
+  end
```
corylown, line 14:
```suggestion
    def icon
      blacklight_icon helpers.document_or_parent_icon(@document)
    end
  end
```

app/components/arclight/search_result_component.html.erb
jcoyne: Ugh, this looks like a bad rebase. Let me straighten it out.

jcoyne: Okay, it should be fixed up now.

corylown (approved):

PR #1289, opened 2022-11-08
Provide a fallback for the logo link

cbeer:
Without this, we can get some pretty weird routing errors.


jcoyne:
@cbeer this is because ostruct isn't raising a `NoMethodException`? Only for Blacklight 7?

cbeer:
I didn't (and still don't?) have `blacklight_config.logo_link` defined, and `link_to 'whatever', nil` tries to route to the current path (and fails miserably in an engine context?).

jcoyne:
@cbeer wouldn't this be giving you the default https://github.com/projectblacklight/blacklight/blob/0bdfc3a503ade8719876abf21901853b7b3b0a69/lib/blacklight/configuration.rb#L351-L353?

cbeer:
Rebuilding the test app fixed this.

PR #1295, opened 2022-11-08
Remove unused field

jcoyne:
We never index names_sim, so remove it from facet.field


cbeer:
I see a `names_ssim`; I wonder if this was just a typo we should fix?

jcoyne:
@cbeer I don't have enough information to know if that would be better.

jcoyne:
Superseded by #1314

PR #1302, opened 2022-11-09
Upgrade the more info button

### LINKED ISSUE #1301: Upgrade the collection info button
Since we relocated the Request button from the Collection box, there is now more horizontal space available in the bottom row that contains the Download and collection info buttons. Especially if the collection has no downloads available, and thus no Download button to display:

Screen Shot 2022-11-08 at 4 51 13 PM

I suggest we take advantage of the available space and increase the clarity of the collection info button by updating its styling to make it a regular-looking button:

Screen Shot 2022-11-08 at 3 00 45 PM

Some considerations:

- The upgraded more info button should use the same styling as the Download button
- Keep the info icon but change its color to `#F5F5F5`
- If there is no download button, the More info button should be left-aligned, the way the Download button is in the mockup above
- Nothing about the More info button's pulldown display of collection attributes needs to change (I don't think)

jcoyne:
Fixes #1301
Screenshot 2022-11-09 at 8 16 09 AM

### DIFF
--- a/app/components/arclight/collection_info_component.html.erb
+++ b/app/components/arclight/collection_info_component.html.erb
@@ -1,8 +1,8 @@
 <div class='al-show-actions-box-info d-inline-flex'>
 
   <div class="dropdown">
-    <button class="dropdown-toggle btn" type="button" data-bs-toggle="dropdown" aria-expanded="false" style="border: 0">
-      <span style="color: gray"><%= info_icon %></span>
+    <button class="dropdown-toggle btn btn-secondary btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+      <%= info_icon %> <%= t('.more_info') %>
     </button>
     <div class="dropdown-menu p-2" style="min-width: 300px;" >
       <table style="width: 100%">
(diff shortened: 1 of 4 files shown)


app/components/arclight/collection_info_component.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class='al-show-actions-box-info d-inline-flex'>
 
   <div class="dropdown">
-    <button class="dropdown-toggle btn" type="button" data-bs-toggle="dropdown" aria-expanded="false" style="border: 0">
-      <span style="color: gray"><%= info_icon %></span>
+    <button class="dropdown-toggle btn btn-secondary btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false" style="border: 0">
```
cbeer, line 4:
```suggestion
    <button class="dropdown-toggle btn btn-secondary btn-sm" type="button" data-bs-toggle="dropdown" aria-expanded="false">
```

Is the border style cruft too?

jcoyne: Yeah. No idea why that was there.

PR #1309, opened 2022-11-10
Pass the component config through the indexing hierarchy

cbeer:


jcoyne:
Looks good, but can you let us know why we need this change?

cbeer:
If you don't, any 2nd (or higher) level child component doesn't use our configured indexing file? 🤷‍♂️

jcoyne:
I'm still not grasping what this is about. Can you perhaps share an example of how this would be used?

cbeer:
EAD is hierarchical. We index the collection using `ead2_config` and any component elements (`c`, `c1`, `c2`, etc) using `ead2_component_config`. Collections and components are indexed hierarchically (so the collection renders any of its direct children, but a `c2` component will render its `c3` children). 

We created a seam for downstream applications to specify which files arclight should use (in #1268), which allows downstream to pass through traject settings to specify which files, however `ead2_compoent_config` continued to hard-code the arclight-provided traject configuration, so we could only use our own configurations for the collection and very top-level children.

This PR makes sure we pass the requested traject config file setting through the whole hierarchy so we use it as one might expect.

PR #1322, opened 2022-11-14
Update the stylesheet to point at the currently use identifier

### LINKED ISSUE #1321: #access-and-use should have the same style as #metadata on the Collection page
it looks like we style `#access` which is no longer in use.

jcoyne:
It was previously changed from #access to #access-and-use

Fixes #1321


marlo-longley:
@jcoyne this looks good; can you rebase?

thatbudakguy (approved):

PR #1324, opened 2022-11-15
Add bit of margin to info button icon

ggeisler:
Super minor visual thing but the collection info button's icon feel a little crowded against the button label:

Screen Shot 2022-11-14 at 5 27 31 PM

This PR just gives it a bit of horizontal separation:

Screen Shot 2022-11-14 at 5 29 52 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/icons.scss
+++ b/app/assets/stylesheets/arclight/modules/icons.scss
@@ -2,3 +2,7 @@
   height: 1rem;
   width: 1rem;
 }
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
This was the unchanged icon from bootstrap.  How about we add this to our local sass instead. I'm thinking something like `.bi-info-circle-fill { @extend .me-2 } `

jcoyne (approved):

PR #1329, opened 2022-11-17
Display the bookmark in the contained components area

### LINKED ISSUE #1267: Update "Contents" section of parent component page
It looks like something that didn't make it over in the recent UI redesign is the bookmarks for items in the Contents section of a parent component page. The current layout right-aligns the Box/Folder info and omits the bookmark controls:

Screen Shot 2022-11-02 at 5 22 33 PM

---

I believe we want something like this, where the bookmark controls are present for each item (right-aligned) and the box/folder info is left-aligned, similar to the search result item layout:

Screen Shot 2022-11-02 at 5 23 31 PM

jcoyne:
This involved splitting the DocumentCollectionHierarchyComponent into 2 separate components, which allows for the removal of a number of conditionals

Fixes #1267

### DIFF
--- a/app/components/arclight/document_collection_context_component.html.erb
+++ b/app/components/arclight/document_collection_context_component.html.erb
@@ -1,4 +1,4 @@
-<%= content_tag :li,
+<%= tag.tr(
   id: @document.id,
   data: {
     'document-id': @document.id.to_s.parameterize,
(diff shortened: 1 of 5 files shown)


thatbudakguy (changes requested):
Putting the bookmark checkboxes in this context seems to have surfaced some additional style bugs for me.

1. bookmark checkboxes are incorrectly sized

Screen Shot 2022-11-30 at 13 29 01

i think being inside a flex context next to the online content icon causes this. the bookmark checkboxes seem to have some styles applied that don't make sense to me (why is their height set to `var(--bs-card-height)`?) setting the height to just `1rem` fixed this.

2. bookmark checkboxes aren't aligned to match online content icon

Screen Shot 2022-11-30 at 13 30 49

this can be easily fixed by making the parent, which has `d-inline-flex justify-content-end`, also have `align-items-center`.

3. top container info is awkwardly wrapped

Screen Shot 2022-11-30 at 13 31 50

gary's design seems to leave a lot more breathing room between the container info, bookmark checkbox, and title. i think we're OK with wrapping the titles rather than wrapping any of the other columns; it looks a little weird to wrap the top container info because it's so short. not a huge deal though.

jcoyne:
>     bookmark checkboxes are incorrectly sized

Those aren't checkboxes. Those are borders around `form.bookmark-toggle`.  I don't know why the bookmarks are not displaying.

thatbudakguy:
that would explain why I also couldn't click them!! thanks

jcoyne:
Screenshot 2022-11-30 at 3 54 00 PM

thatbudakguy (approved):

PR #1339, opened 2022-11-22
Highlight the collection context link when it is active

### LINKED ISSUE #1280: Add background highlighting to Collection overview section when active
Currently when the user selects a node from the Collection contents tree, the component page for that node is displayed in the main content area and background highlighting is added to that node in the Collection contents tree. This highlighting helps the user understand what component is currently selected, and reinforces the association between the sidebar navigation element and its associated content in the main content area:

Screen Shot 2022-11-07 at 2 17 54 PM

---

We should do the same thing for the Collection overview section, for the same reasons:

Screen Shot 2022-11-07 at 2 37 26 PM

---

Some implications of this:

- When a collection page is first loaded (via a breadcrumb link, search result link, etc.), the first item in the collection overview nav should be highlighted.
- Anytime the collection sidebar is displayed, one (and only one) thing in it should be highlighted: either a nav item in Collection overview (when the Collection page is displayed in the main content area) or a node in the Collection contents tree (when a component is displayed in the main content area).
- The combination of the two points above are important in helping the user learn how the sidebar works in relation to the main content area, which can be a little non-intuitive at first because of how the collection and component details look very different in the main content area.

jcoyne:
Fixes #1280


thatbudakguy:
moving this to draft pending discussion of requirements for #1280

PR #1346, opened 2022-11-28
Remove extra margin from breadcrumbs

### LINKED ISSUE #1342: Reduce bottom margin on breadcrumb section
It looks like there is bottom margin applied to the `breadcrumb` class, but that class can show up at least twice (the second nested within the first), so we get too much bottom margin in total:

Screen Shot 2022-11-23 at 4 53 56 PM

---

I think the spacing would be about right if we only applied the bottom margin from one of these (e.g., maybe just the outer one, but I haven't had time to experiment). It basically feels twice as much as we'd ideally want.

thatbudakguy:
Fixes #1342

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -6,13 +6,15 @@
 
 .al-show-breadcrumb {
   border-bottom: $border-width solid $border-color;
-  padding-bottom: $spacer * .5;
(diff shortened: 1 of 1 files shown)


thatbudakguy:
screenshots showing the change, including with padding/margin inspector active:

Screen Shot 2022-11-28 at 14 39 43

Screen Shot 2022-11-28 at 14 39 38

jcoyne (approved):

PR #1347, opened 2022-11-28
Ensure last breadcrumb doesn't break onto new line

### LINKED ISSUE #1325: Terminal node of breadcrumb shouldn't be on new line
This isn't a super high priority issue, but the breadcrumb structure is still slightly off from the intended design.

The terminal node - the current page title - shouldn't break to a new line unless it has to because there isn't enough horizontal space in the current breadcrumb line to fit the entire node. 

It appears we are always breaking the terminal node to its own line, even when there is plenty of space left on the previous line:

Screen Shot 2022-11-15 at 9 22 12 AM

In that example, the " >> Item AA001" node should follow directly after its parent node, like "File 1 >> Item AA001", since there's sufficient horizontal space for it to completely fit.

(The "File 1" node is correctly on a new line, because we want the collection node to always be on its own line to better highlight it is the parent collection.)

thatbudakguy:
Fixes #1325

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -27,6 +27,11 @@
         content: escape-svg(url($bent-arrow));
       }
     }
+
+    // last item doesn't get its own line
+    .breadcrumb-item-4 {
+      flex: none;
+    }
   }
 
   .breadcrumb-item-4 {


jcoyne (changes requested):
Can you add screenshots showing the impact of this change for both a collection and a lower level component?

thatbudakguy:
collection:

Screen Shot 2022-11-28 at 16 06 34

item with a long title (wraps to its own line):

Screen Shot 2022-11-28 at 16 08 21

item with a short title (same line):

Screen Shot 2022-11-28 at 16 08 39

jcoyne (approved):
Thanks.

PR #1348, opened 2022-11-29
Update solr schema to play nice with solr atomic update

cbeer:
This PR ensures that all document fields coming out of the indexer are stored (so can be appropriately regenerated by Solr), and that copyField destinations are not (so Solr doesn't double up on them)

I tried to collapse indexed fields where they are obviously the same, but there's still some redundancy (e.g. `creator_ssm`, `creator_ssim`, and `creators_ssm` sure seem to be the same 🤷‍♂️) that we might want to clean up in the future.

There's also quite a few indexed fields that seem entirely unused (`title_filing_ssi`, etc).

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -280,9 +280,9 @@
    <field name="id" type="string" indexed="true" stored="true" required="true" />
    <field name="_version_" type="plong" indexed="true" stored="true" multiValued="false" />
    <field name="timestamp" type="pdate" indexed="true" stored="true" default="NOW" multiValued="false"/>
-   <!-- default, catch all search field -->
+   <!-- default, catch all search field, also used for search term highlighting (requires stored="true") -->
    <field name="text" type="text" indexed="true" stored="true" multiValued="true"/>
-   <field name="unitid_identifier_match" type="identifier_match" indexed="true" multiValued="true" />
+   <field name="unitid_identifier_match" type="identifier_match" indexed="true" stored="false" multiValued="true" />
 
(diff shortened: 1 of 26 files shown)


jcoyne:
Can you explain why we want to have Atomic updates? Is there a place where we can document this decision so that future changes don't unknowingly depart from this path?

jcoyne:
> Where do you think this should be documented?

Either in the Traject config or the dynamicField definition section of the schema.xml (https://github.com/projectblacklight/arclight/blob/main/solr/conf/schema.xml#L300-L319)

solr/conf/schema.xml
```diff
@@ -317,6 +317,7 @@
    <dynamicField name="*_isim"  type="pint"      stored="true"  indexed="true"  multiValued="true"  />
    <dynamicField name="*_is"    type="pint"      stored="true"  indexed="false" multiValued="false" />
    <dynamicField name="*_ii"    type="pint"      stored="false" indexed="true"  multiValued="false" />
```
jcoyne, line 319: Should we remove all the non-stored dynamic fields?  It seems like leaving these in could get people into a situation where they can't do atomic-updates.

jcoyne (changes requested):

solr/conf/schema.xml
cbeer: 🤷‍♂️ we use some of them in copy fields, and I figured it might be nice for downstream not to throw them away immediately.

jcoyne (approved):

PR #1354, opened 2022-12-01
Prevent collection info dropdown from being cut off

thatbudakguy:
Fixes https://github.com/sul-dlss/vt-arclight/issues/283


thatbudakguy:
Screen Shot 2022-12-01 at 10 07 59

jcoyne (approved):

PR #1369, opened 2022-12-15
Allow the highlighted field to be configurable

### LINKED ISSUE #1368: Highlight field to be configurable
Currently it's hardcoded as `text`
https://github.com/projectblacklight/arclight/blob/82d961727e685f99f44daad5765fea1b29ad4ea3/app/models/concerns/arclight/solr_document.rb#L123-L125

https://github.com/projectblacklight/arclight/blob/251842d0a227ca968eaa32a6f1c215f31c279f3f/app/models/concerns/arclight/search_behavior.rb#L34

jcoyne:
Fixes #1368

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -25,6 +25,9 @@ class CatalogController < ApplicationController
       'collection.rows': 1
     }
 
+    # Sets the indexed Solr field that will display with highlighted matches
+    config.highlight_field = 'text'
+
     # solr path which will be added to solr base url before the other solr params.
     # config.solr_path = 'select'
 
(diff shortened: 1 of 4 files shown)


lib/generators/arclight/templates/catalog_controller.rb
```diff
@@ -25,6 +25,8 @@ class CatalogController < ApplicationController
       'collection.rows': 1
     }
 
+    config.highlight_field = 'text'
```
marlo-longley, line 28: @jcoyne it may be obvious but could we put a comment that is like, this sets which indexed field will display with highlighted matches per search ?

jcoyne: I think this is a great idea. Will do.

marlo-longley (approved):

PR #1376, opened 2022-12-20
Update attributes to use new Blacklight syntax

jcoyne:
A default of `nil` is now required unless the field will always have a value.
See https://github.com/projectblacklight/blacklight/pull/2804


app/models/concerns/arclight/solr_document.rb
```diff
@@ -54,7 +54,8 @@ def repository
     end
 
     def repository_and_unitid
-      [repository, unitid].compact.join(': ')
+      # We need compact_blank here due to https://github.com/projectblacklight/blacklight/issues/2952
+      [repository, unitid].compact_blank.join(': ')
```
corylown, line 58: Is `compact_blank` still needed now that this is fixed: https://github.com/projectblacklight/blacklight/issues/2952

corylown (approved):

PR #1379, opened 2022-12-20
Support passing extra params to the oEmbed viewer

### LINKED ISSUE #1378: Support passing extra params to oEmbed providers
The current implementation for the oEmbed viewer visits the URL provided by the `data-arclight-oembed-url` attribute, looks for a `<link rel="alternate" type="application/json+oembed">` on that page, and fetches the oEmbed information from that `<link>`.

This works great if the object you're embedding has a page that provides its oEmbed URL, but makes it tricky to pass extra parameters to the viewer because the object's page has to also support those params.

For example:
* an item has the unique identifier `foobar`
* the permalink for that item is `https://purl.myschool.edu/foobar`
* the permalink page provides an oEmbed URL in a `<link>` in its `<head>`: `https://embed.myschool.edu/embed.json?url=https%3A%2F%2Fpurl.myschool.edu%2Ffoobar`
* we want to embed an oEmbed viewer and pass the parameter `feature=baz` to it, which will enable some feature in the viewer. this can be done with `https://embed.myschool.edu/embed.json?feature=baz&url=https%3A%2F%2Fpurl.myschool.edu%2Ffoobar`

The page `https://purl.myschool.edu/foobar` doesn't recognize any such parameter `feature=`, and will always provide `https://embed.myschool.edu/embed.json?url=https%3A%2F%2Fpurl.myschool.edu%2Ffoobar` as the oEmbed URL, preventing us from forwarding this param on to the viewer.

In this scenario, our best option may be to follow the `<link>` to get the oEmbed URL, then modify it after-the-fact to add our own parameters to pass to the viewer. These parameters could be stored as additional `data-` attributes on the viewer template, e.g.
```html
<div
  data-arclight-oembed="true"
  data-arclight-oembed-url="https://purl.myschool.edu/foobar"
  data-arclight-oembed-feature="baz"> // arbitrary params at data-arclight-oembed-*
</div>
```

thatbudakguy:
Closes #1378

Supports https://github.com/sul-dlss/vt-arclight/issues/155

If you add, for example, `data-arclight-oembed-search="hess"` and `data-arclight-oembed-hide_title="true"` to the viewer's template (two parameters supported by sul-embed), you now get:

Screen Shot 2022-12-20 at 11 50 10

### DIFF
--- a/app/assets/javascripts/arclight/oembed_viewer.js
+++ b/app/assets/javascripts/arclight/oembed_viewer.js
@@ -5,27 +5,52 @@ class OembedLoader {
 
   load() {
     const loadedAttr = this.el.getAttribute('loaded')
-    var data = this.el.dataset
-    var resourceUrl = data.arclightOembedUrl
+    const { arclightOembed, arclightOembedUrl, ...extraOembedParams } = this.el.dataset
+    const extraParams = OembedLoader.normalizeParams(extraOembedParams)
+
     if (loadedAttr && loadedAttr === 'loaded') {
       return
     }
 
-    fetch(resourceUrl)
+    fetch(arclightOembedUrl)
       .then((response) => response.text())
       .then((body) => {
-        const oEmbedEndPoint = OembedLoader.findOEmbedEndPoint(body)
+        const oEmbedEndPoint = OembedLoader.findOEmbedEndPoint(body, extraParams)
         if (!oEmbedEndPoint || oEmbedEndPoint.length === 0) {
+          console.warn(`No oEmbed endpoint found in <head> at ${arclightOembedUrl}`)
           return
         }
         this.loadEndPoint(oEmbedEndPoint)
       })
   }
 
-  static findOEmbedEndPoint(body) {
+  // Convert data-arclight-oembed-* attributes to URL parameters for the viewer
+  static normalizeParams(attributes) {
+    return Object.keys(attributes).reduce((acc, attribute) => {
+      // Reverse data attribute name conversion. See:
+      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+      const parameterName = attribute.replace('arclightOembed', '')
+        .replace(/([a-z])([A-Z])/g, '$1-$2')
+        .toLowerCase()
+
+      acc[parameterName] = attributes[attribute]
+      return acc
+    }, {})
+  }
+
+  static findOEmbedEndPoint(body, extraParams = {}) {
+    // Parse out link elements so image assets are not loaded
     const template = document.createElement('template')
-    template.innerHTML = body.match(/<link .*>/g).join('') // Parse out link elements so image assets are not loaded
-    return template.content.querySelector('link[rel="alternate"][type="application/json+oembed"]')?.getAttribute('href')
+    template.innerHTML = body.match(/<link .*>/g).join('')
+
+    // Look for a link element containing the oEmbed endpoint; bail out if none
+    const endpoint = template.content.querySelector('link[rel="alternate"][type="application/json+oembed"]')
+      ?.getAttribute('href')
+    if (!endpoint) return ''
+
+    // Serialize any extra params and append them to the endpoint
+    const qs = new URLSearchParams(extraParams).toString()
+    return `${endpoint}&${qs}`
   }
 
   loadEndPoint(oEmbedEndPoint) {


app/assets/javascripts/arclight/oembed_viewer.js
```diff
@@ -5,27 +5,45 @@ class OembedLoader {
...
+        const oEmbedEndPoint = OembedLoader.findOEmbedEndPoint(body, extraParams)
         if (!oEmbedEndPoint || oEmbedEndPoint.length === 0) {
           return
         }
         this.loadEndPoint(oEmbedEndPoint)
       })
   }
 
-  static findOEmbedEndPoint(body) {
+  // Convert data-arclight-oembed-* attributes to params for the viewer
```
jcoyne, line 26:
```suggestion
  // Convert data-arclight-oembed-* attributes to URL parameters for the viewer
```

app/assets/javascripts/arclight/oembed_viewer.js
```diff
@@ -5,27 +5,45 @@ class OembedLoader {
...
+  static normalizeParams(params) {
+    return Object.keys(params).reduce((acc, key) => {
+      // Reverse data attribute name conversion. See:
+      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+      const newKey = key.replace('arclightOembed', '')
+        .replace(/([a-z])([A-Z])/g, match => `${match[0]}-${match[1].toLowerCase()}`)
+        .toLowerCase()
+
+      acc[newKey] = params[key]
```
jcoyne, line 35:
```suggestion
  static normalizeParams(attributes) {
    return Object.keys(attributes).reduce((acc, attribute) => {
      // Reverse data attribute name conversion. See:
      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
      const parameterName = attribute.replace('arclightOembed', '')
        .replace(/([a-z])([A-Z])/g, match => `${match[0]}-${match[1].toLowerCase()}`)
        .toLowerCase()

      acc[parameterName] = attributes[attribute]
```

app/assets/javascripts/arclight/oembed_viewer.js
```diff
@@ -5,27 +5,45 @@ class OembedLoader {
...
 
-  static findOEmbedEndPoint(body) {
+  // Convert data-arclight-oembed-* attributes to params for the viewer
+  static normalizeParams(params) {
+    return Object.keys(params).reduce((acc, key) => {
+      // Reverse data attribute name conversion. See:
+      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+      const newKey = key.replace('arclightOembed', '')
+        .replace(/([a-z])([A-Z])/g, match => `${match[0]}-${match[1].toLowerCase()}`)
+        .toLowerCase()
```
jcoyne, line 33: Do we need two calls to `toLowerCase()`?

jcoyne (changes requested):

app/assets/javascripts/arclight/oembed_viewer.js
thatbudakguy: The second one handles the case where there is only one word, e.g. `search`. It'll still be capitalized since the original attribute name was `arclightOembedSearch`, so we need to lowercase it after removing `arclightOembed`.

jcoyne: Do we need the first one then?

thatbudakguy: I guess not! makes the replacement simpler too.

thatbudakguy: Fixed.

jcoyne (approved):

thatbudakguy:
Do you know why CI is failing? It seems like some items aren't being indexed

marlo-longley:
Agreed, indexing is not working. It's looking for catalog/undefined and I'm not sure if undefined is coming from issues with indexing or somewhere in the app code. It seems like indexing worked, in the logs of the test run.

Line 1298 of the test logs:
```
2023-01-05 22:29:00 +0000 Rack app ("GET /catalog/undefined&" - (127.0.0.1)): #<Blacklight::Exceptions::RecordNotFound: Blacklight::Exceptions::RecordNotFound>
2023-01-05 22:29:00 +0000 Rack app ("GET /catalog/undefined&" - (127.0.0.1)): #<Blacklight::Exceptions::RecordNotFound: Blacklight::Exceptions::RecordNotFound>
2023-01-05 22:29:00 +0000 Rack app ("GET /catalog/undefined&" - (127.0.0.1)): #<Blacklight::Exceptions::RecordNotFound: Blacklight::Exceptions::RecordNotFound>
2023-01-05 22:29:01 +0000 Rack app ("GET /catalog/undefined&" - (127.0.0.1)): #<Blacklight::Exceptions::RecordNotFound: Blacklight::Exceptions::RecordNotFound>
2023-01-05 22:29:17 +0000 Rack app ("GET /catalog/undefined&" - (127.0.0.1)): #<Blacklight::Exceptions::RecordNotFound: Blacklight::Exceptions::RecordNotFound>
```
(The warnings about missing IDs below 1298 is expected; see https://github.com/projectblacklight/arclight/issues/785)

thatbudakguy:
Figured this one out — fixture object aspace_843e8f9f22bac69872d0802d6fffbb04 includes one `<dao>` that's supposed to embed a viewer, but [the page it goes to](https://collections.nlm.nih.gov/bookviewer?PID=nlm:nlmuid-100957835-bk) doesn't actually have a `link[rel="alternate"][type="application/json+oembed"]` pointing to an oEmbed URL.

I added a check to skip trying to embed it if this happens, and also output a warning to the browser console that we tried to embed something that was not oEmbeddable.

PR #1380, opened 2023-01-06
Use dt and dd elements inside result metadata container

thatbudakguy:
We use a dl for search result metadata display, but don't use the
corresponding child elements to mark up the actual metadata
fields and values. This changes the elements and inserts the
labels for metadata fields as visually-hidden dt so that screen
reader users can access them.

Ref https://github.com/sul-dlss/vt-arclight/pull/444


thatbudakguy:
blocked by #1381

marlo-longley:
@thatbudakguy https://github.com/projectblacklight/arclight/pull/1381 merged !

marlo-longley (approved):

PR #1382, opened 2023-01-09
Only apply card styling to bookmark toggle in title container

thatbudakguy:
This ensures that lists of child components are more concise
and readable by removing the card padding/styling. The card style
is used only for the main bookmark toggle in the title/header.

Fixes https://github.com/sul-dlss/vt-arclight/issues/448


thatbudakguy:
Screen Shot 2023-01-09 at 12 42 26

jcoyne (approved):

ggeisler (approved):
👍

PR #1387, opened 2023-01-13
Add drop shadow to Collection info dropdown

ggeisler:
This is a small thing but the Collection info dropdown box can kind of blend in with the underlying page if the user isn't paying attention when opening it. This PR adds a drop shadow to the box to make it stand out from the underlying page a bit.

### Before
Screen Shot 2023-01-13 at 1 59 30 PM

### After
Screen Shot 2023-01-13 at 2 04 38 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/layout.scss
+++ b/app/assets/stylesheets/arclight/modules/layout.scss
@@ -123,6 +123,7 @@
     .dropdown-toggle:after {
       margin-left: 0;
     }
+
     table {
       td {
         text-align: right;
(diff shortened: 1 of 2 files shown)


cbeer:
Could we just add the `shadow` class to get the same effect?

https://getbootstrap.com/docs/5.3/utilities/shadows/#examples

ggeisler:
> Could we just add the shadow class to get the same effect?

Probably! I didn't realize, or more likely forgot, that Bootstrap offered shadows.

I'll revise this PR next week; I've got VT running now and don't want to switch and have to do all the indexing again, etc.

ggeisler:
Updated to use Bootstrap shadow class instead of manually created shadow:

### Before
Screen Shot 2023-01-17 at 9 25 33 AM

### After
Screen Shot 2023-01-17 at 9 31 08 AM

marlo-longley (approved):

PR #1388, opened 2023-01-18
Color contrast improvements for accessibility

ggeisler:
A couple of minor color contrast updates for things that didn't have sufficient contrast for accessibility purposes:

Links don't have sufficient contrast against the repository box background color. No lighter gray, in fact, provides sufficient contrast. So I went with removing the background color. We can assume implementers will style these differently to match their local style, anyway.

### Before
Screen Shot 2023-01-17 at 5 26 11 PM

### After
Screen Shot 2023-01-17 at 4 19 06 PM

---

Same thing with the online content box. No lighter version of green provides sufficient contrast with the default Bootstrap blue link. Basically same solution as above.

### Before
Screen Shot 2023-01-17 at 5 25 16 PM

### After
Screen Shot 2023-01-17 at 5 24 59 PM

### DIFF
--- a/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
@@ -141,6 +141,7 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
 
   .online-contents {
     grid-area: online-contents;
+    margin-bottom: 1rem;
   
     h2 {
       font-size: 1rem;
--- a/app/assets/stylesheets/arclight/modules/repository_card.scss
+++ b/app/assets/stylesheets/arclight/modules/repository_card.scss
@@ -1,6 +1,5 @@
 .al-repository {
   border: $default-border-styling;
-  background-color: $gray-200;
 }
 
 @include media-breakpoint-up(sm) {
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
```diff
@@ -141,6 +141,10 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
 
   .online-contents {
     grid-area: online-contents;
+
+    .alert {
```
jcoyne, line 145: This to me seems like we're using alert improperly.  These are cards, not alerts.  Can we remove the class from the template instead?

jcoyne (changes requested):

app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
jcoyne, line 145: Can we style `.alert-dark` instead here?

jcoyne (changes requested):

ggeisler:
I've updated the online content box to use `card` rather than `alert`. 

(I also added a hidden span for the box title because I noticed in some other accessibility testing that we have duplicate headings with the text "Online content". The hidden text will help screenreader users by adding "Filter" to the first heading's text and remove the duplication.)

It looks pretty much like it did previously:

Screen Shot 2023-01-18 at 1 35 51 PM

jcoyne (approved):
Thanks. This is much clearer.

marlo-longley (approved):

PR #1389, opened 2023-01-18
Improve results when searching for unitids

marlo-longley:
- Remove duplicate WordDelimiterGraphFilterFactory on unitid field
- Boost relevance of unitid_identifier_match in search results

This work comes out of observations made in https://github.com/sul-dlss/vt-arclight/issues/186 -- when searching by unit identifier, the corresponding item came multiple pages deep in the search results.

### DIFF
--- a/solr/conf/solrconfig.xml
+++ b/solr/conf/solrconfig.xml
@@ -89,9 +89,9 @@
          collection_title_tesim^150
          title_tesim^100
          normalized_title_teim^100
+         unitid_identifier_match^40
          name_teim^10
          place_teim^10
-         unitid_identifier_match^5
          subject_teim^2
          text
        </str>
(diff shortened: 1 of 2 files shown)


solr/conf/solrconfig.xml
```diff
@@ -91,7 +91,7 @@
          normalized_title_teim^100
          name_teim^10
          place_teim^10
-         unitid_identifier_match^5
```
thatbudakguy, line 94: small thing, but could we bump this up above name/place? it's nice to have them sorted by boost in descending order

thatbudakguy (approved):

solr/conf/solrconfig.xml
thatbudakguy: to clarify: just shift the line upwards, but keep the same boosting, so that the boost fields are sorted

marlo-longley: @thatbudakguy good idea, done!

PR #1396, opened 2023-02-01
Test against a known blacklight release

cbeer:


jcoyne (approved):

jcoyne (changes requested):
This causes the non-matrix build to run with BL 7.33, which will fail.

jcoyne (approved):

marlo-longley:
The BL commit that the AL tests start failing on: https://github.com/projectblacklight/blacklight/commit/3e4a46f19bdb285af3ad75e43c56de85854a67ad. This commit was included in beta5.
Before that, all the way down to BL beta1, the tests in AL pass.

PR #1403, opened 2023-02-24
Test on concurrent-ruby 1.2.0

jcoyne:


jcoyne:
Clearly, this was not what caused the problem.

Issue #1401, opened 2023-02-24
Blacklight is not defined error in javascript


jcoyne:
<img width="1707" alt="Screenshot 2023-02-24 at 1 04 51 PM" src="https://user-images.githubusercontent.com/92044/221269104-379a794b-15d2-4bad-b69d-7f7f6756a6fa.png">

PR #1409, opened 2023-02-27
Use urlValue rather than urlTarget for the url

jcoyne:
Also added error handling in case the URL doesn't resolve

### DIFF
--- a/app/assets/javascripts/arclight/oembed_controller.js
+++ b/app/assets/javascripts/arclight/oembed_controller.js
@@ -1,6 +1,6 @@
 import { Controller } from '@hotwired/stimulus'
 
-export default class extends Controller {
+export default class OembedController extends Controller {
   static values = {
     url: String
   }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/arclight/oembed_controller.js
```diff
@@ -12,32 +12,23 @@ export default class extends Controller {
...
-      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
-      const parameterName = attribute.replace('arclightOembed', '')
-        .replace(/([a-z])([A-Z])/g, '$1-$2')
-        .toLowerCase()
-
-      acc[parameterName] = attributes[attribute]
-      return acc
-    }, {})
-  }
-
```
thatbudakguy, line 40: why are we getting rid of this? we need it to parse the extra params, which is what powers in-document search in VT: https://github.com/sul-dlss/vt-arclight/pull/451

did we change the implementation of this to drop support for passing extra params to the viewer?

jcoyne: It didn't look like this did anything. Maybe it needs a test?

thatbudakguy: Why did support go away? Does stimulus prevent us from using arbitrary `data-*` attributes?

If we need this PR in order to fix anything in VT, we should be aware that upgrading our arclight version will break in-document search until and unless we reimplement it there.

jcoyne: It went away because of the need to fix https://github.com/projectblacklight/arclight/issues/1401 and because it wasn't clear to me that it was used for anything (not tested and the documentation was unclear about the need for it.)

thatbudakguy (approved):

Issue #1391, opened 2023-01-19
Duplicate id values on parent container component page causes a11y issue


ggeisler:
I'm not sure how easy this will be to resolve but to record the issue: on the /catalog page where we display the collection sidebar with the collection contents tree, if the current item is a parent container component, we have a duplicate id value, which creates an accessibility issue. 

For example, in this example (`/catalog/umich-bhl-851981aspace_4c08111acad2d66e8a63404ad6ce4b46`), the two areas outlined in red are turbo-frame elements that both have `id="al-hierarchy-umich-bhl-851981aspace_4c08111acad2d66e8a63404ad6ce4b46"`.

<img width="937" alt="Screen Shot 2023-01-18 at 5 02 58 PM" src="https://user-images.githubusercontent.com/101482/213329539-9305f809-6ac4-47d2-b03d-6cec2108b3ac.png">

According to accessibility guidelines, this is reported as an error because it can potentially cause odd behavior for screenreaders, though I haven't been able to determine exactly what behavior this will produce. But in any case, I can't try all screenreaders (e.g. on Windows) so the ideal resolution is to find another way to represent the two turboframe sections in my screenshot above such that they don't use duplicate id values.

dinahhandel:
Is ID critical to collapse/expand? Append one to ensure they are unique and don't conflict with screen readers.

lfarrell:
Fix PR: https://github.com/projectblacklight/arclight/pull/1435

PR #1435, opened 2023-12-05
Fix issues with duplicate item ids

lfarrell:
Fix issues with duplicate item ids between the collection sidebar and the collection contents display.

Issue: https://github.com/projectblacklight/arclight/issues/1391


randalldfloyd (approved):
Verified with [Siteimprove accessibility checker plugin](https://www.siteimprove.com/integrations/browser-extensions/) that the issue is resolved when running this branch locally.

PR #1437, opened 2023-12-05
Fix template with generator 

### LINKED ISSUE #1436: The Arclight template install method currently not working
Using the application template to install Arclight, per the documentation, fails with:
```
cannot load such file -- sassc-rails (LoadError)
cannot load such file -- dartsass-sprockets (LoadError)
```

This is being caused by a problem with the Blacklight generators as [described in this issue](https://github.com/projectblacklight/blacklight/issues/3031).

In a nutshell:

- A Blacklight generator related to assets gets triggered during the `bundle install` step from within the Arclight template
- The BL generator tries to update the Rails `importmap` configuration.
- The BL generator fails because `importmap:install` has yet to be run by the Rails app generators.
- Since the BL generator fails it doesn't inject `sassc-rails` into the Gemfile in a later step
- The main BL install generator triggered later in the Arclight template fails due to the missing dependency

In contrast, the Arclight application template does work if the Rails application is generated first, and then the template is applied.  This is because the Rails `importmap:install` step has completed before the template steps trigger the Blacklight install generator:

```
rails new arclight_app

cd ./arclight_app

bundle exec rails app:template LOCATION=https://raw.githubusercontent.com/projectblacklight/arclight/main/template.rb
```

marlo-longley:
Closes #1436 

Thanks to @randalldfloyd and @jcoyne for their help with this!

### DIFF
--- a/lib/generators/arclight/install_generator.rb
+++ b/lib/generators/arclight/install_generator.rb
@@ -22,7 +22,7 @@ module Arclight
     def install_blacklight_locale_picker
       gem 'blacklight-locale_picker'
 
-      Bundler.with_clean_env do
+      Bundler.with_unbundled_env do
         run 'bundle install'
       end
 
--- a/template.rb
+++ b/template.rb
@@ -2,9 +2,9 @@
 
 gem 'arclight'
 
-run 'bundle install'
+after_bundle do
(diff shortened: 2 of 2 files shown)


randalldfloyd (approved):
This works for me locally

Issue #1312, opened 2022-11-10
Make collection sidebar sticky and scrollable


ggeisler:
I'd consider this a UX enhancement more than a bug fix because the collection/component pages currently work acceptably in terms of scrolling, but if and when someone has time I think it's a better user experience when the sidebar remains in-place when the main content area is scrolled. This keeps the sidebar in view and accessible when the user scrolls down longer collection/component pages. Duke's implementation does this and is just what we'd ideally want in core, I believe.

<img width="1262" alt="Screen Shot 2022-11-10 at 2 55 13 PM" src="https://user-images.githubusercontent.com/101482/201214114-5eaba1f1-d010-41ef-9e44-30ea55e8d1d1.png">

You can't see the sidebar scrollbar in my screenshot above but part of this solution is that the sidebar needs to be scrollable itself, since a content inventory can taller than the content in the collection/component show page.

This only applies to viewports at `lg` and above, since below `lg` we now collapse the sidebar and display it as an off-canvas panel.

@seanaery One question about your implementation of this: I noticed that you don't use the sticky sidebar until you get to the `xl` viewport width. So the `lg` viewport width has a different behavior (visible sidebar that scrolls with the main content area) than below `lg` (sidebar collapsed) and above `lg` (sticky sidebar). Just curious what the reason for that was, in case it's relevant to how we should replicate your approach in core? (And any thoughts you have about implementing this in core, if you have them -  thanks!)

seanaery:
@ggeisler That's a great catch -- turns out to be just a mistake/oversight on our part. [I'll get that fixed in our app](https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/merge_requests/316/diffs). Overall, I'd say we really like having the collection/nav context stick and think it enhances the UX. We valued that kind of thing even in our pre-ArcLight finding aids application ([2016 blog post](https://blogs.library.duke.edu/bitstreams/2016/02/24/perplexed-by-context-slick-sticky-titles-skip-the-toll-of-the-scroll/)). It used to be challenging to pull off across browsers, but these days with `position: sticky` support and/or Bootstrap `.position-sticky` class, it doesn't take too much fiddling to get it right.

lfarrell:
PR to add this functionality: https://github.com/projectblacklight/arclight/pull/1438

PR #1438, opened 2023-12-05
Draft: Make sidebar hierarchy sticky

lfarrell:
PR for: https://github.com/projectblacklight/arclight/issues/1312

It keeps the collection info fixed while allowing the hierarchy to be scrolled.


lfarrell:
I'm not sure this is a good idea for core as most implementers are likely to customize the header, possibly changing its height. Thus likely throwing off the alignment of a sticky sidebar.

PR #1442, opened 2023-12-06
Ensures the display of containers only changes the case of first character.

### LINKED ISSUE #1433: Containers display should only capitalize the first character without changing the case of subsequent characters
When containers get displayed, ArcLight core intends to capitalize them, e.g., where `box 1` may be indexed, it should display as `Box 1`. The current solution works OK for containers like that. It uses the ruby `.capitalize` method. 
https://github.com/projectblacklight/arclight/blob/main/app/models/concerns/arclight/solr_document.rb#L113-L116

But `.capitalize` is problematic for containers like `box MN6` because it turns them into `Box mn6`; it lowercases all characters after the first one.

randalldfloyd:
Closes #1433

Changes the display of containers, which currently uses `.capitalize`, to only change the first character but leave subsequent characters unchanged.

### DIFF
--- a/app/models/concerns/arclight/solr_document.rb
+++ b/app/models/concerns/arclight/solr_document.rb
@@ -111,8 +111,13 @@ module Arclight
     end
 
     def containers
-      # NOTE: that .titlecase strips punctuation, like hyphens, we want to keep
-      fetch('containers_ssim', []).map(&:capitalize)
+      # NOTE: Keep uppercase characters if present, but upcase the first if not already
+      containers_field = fetch('containers_ssim', []).reject(&:empty?)
+      return [] if containers_field.blank?
+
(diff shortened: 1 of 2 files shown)


seanaery (changes requested):
Thanks for working on this @randalldfloyd . My apologies -- I see the example customization I had shared before was missing a corresponding rspec test. Can you modify/supplement the spec for `#containers` here as part of this PR to show different cases in action?:
https://github.com/projectblacklight/arclight/blob/main/spec/models/concerns/arclight/solr_document_spec.rb#L90-L96

randalldfloyd:
@seanaery Thanks! Yes I also got to thinking that I should add a spec for it after I took the draft status off of it.

marlo-longley (approved):

Issue #989, opened 2020-02-18
Multiple extents always expressed as consecutive but sometimes concurrent


jlhardes:
Extent seems to always be showing as a consecutive list, comma-separated and with "and" before the last item. If multiple extents are within a single physdesc in the EAD, the same measurement is being expressed in multiple ways:

<img width="498" alt="Screen Shot 2020-02-18 at 9 55 04 AM" src="https://user-images.githubusercontent.com/3440166/74747414-ea3f1080-5234-11ea-8e2d-b36a8a3d6075.png">

That should show in the finding aid display as a single extent:

224 items (1 box)

But it is showing with an "and" like these are two separate physical items: 224 items and (1 box)

Extents that express consecutive lists of items should be in separate physdesc elements in the EAD:

<img width="548" alt="Screen Shot 2020-02-18 at 9 56 52 AM" src="https://user-images.githubusercontent.com/3440166/74747475-0478ee80-5235-11ea-8f9f-4a62a7a282de.png">

That extent should show as ArcLight is showing now:

1 document case, 16 DVDRs, 25 CDRs, and 6 audiocassettes

So there seems to be something that understands to enclose multiple extents from a single physdesc in parentheses but there is something that is formatting multiple extents from the same physdesc as a list instead of formatting them as a single extent.

Issue #997, opened 2020-07-13
Allow display of multiple extents


eliotjordan:
We have a use case to display multiple collection-level extents. Proposal is to render badges for each extent value instead of plucking the first:

https://github.com/projectblacklight/arclight/blob/master/app/models/concerns/arclight/solr_document.rb#L57

![multi-extent](https://user-images.githubusercontent.com/784196/87350775-f5654680-c51d-11ea-9adb-3bfb4d3e8bf1.png)

seanaery:
@eliotjordan We encountered this problem at Duke as well, so I second the desire to fix it in core. Rather than put each extent into its own badge, our solution (which may not be desired for everyone) was to concatenate / delimit them with a dash within one badge, then limit the max-width of the badge via CSS. A couple related changes we made to extent:
- Include all values from the physdesc field, not just extent
- Singularize values like "1 boxes" which appear frequently in ASpace exports
https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/blob/develop/app/models/solr_document.rb#L95-104

andyzyl321:
> @eliotjordan We encountered this problem at Duke as well, so I second the desire to fix it in core. Rather than put each extent into its own badge, our solution (which may not be desired for everyone) was to concatenate / delimit them with a dash within one badge, then limit the max-width of the badge via CSS. A couple related changes we made to extent:
> 
> * Include all values from the physdesc field, not just extent
> * Singularize values like "1 boxes" which appear frequently in ASpace exports
>   https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/blob/develop/app/models/solr_document.rb#L95-104

@seanaery Hello Sean, I'm a student from Indiana University and currently working with <extent> problem on Arclight. Specifically, all <extent> values in collection/component level will be connected by comma, with an 'and' before the last value. For instance, the <physdesc> tag in xml file is showing up below:

<img width="1157" alt="Screen Shot 2020-08-10 at 9 59 03 AM" src="https://user-images.githubusercontent.com/44649202/89790774-6b92a400-daf0-11ea-8f06-245f526def68.png">

It is showing as this in Arclight: 
**3 cubic feet, One cubic-foot records case; one letter-size documents case; and twenty-six shelved books., and 4 CDs** 
The way we want it to show up is: 
**3 cubic feet (One cubic-foot records case; one letter-size documents case; and twenty-six shelved books. and 4 CDs)**
[...]

PR #1446, opened 2023-12-07
Allow for multiple extents

marlo-longley:
I would love someone to verify that I interpreted the intent written in this ticket correctly:
Closes https://github.com/projectblacklight/arclight/issues/989

I would also love feedback on whether we need to repeat this logic for both the `_tesim` and `ssm` fields.

Also, since there was no discussion on the ticket, I assume this is non-controversial and desired by the group.

### DIFF
--- a/app/components/arclight/group_component.html.erb
+++ b/app/components/arclight/group_component.html.erb
@@ -7,7 +7,9 @@
         </div>
       <% end %>
       <h3><%= helpers.link_to_document document %></h3>
-      <%= tag.span(document.extent, class: 'al-document-extent badge') if document.extent %>
+      <% document.extent.each do |extent| %>
+        <%= tag.span extent, class: 'al-document-extent badge' unless compact? %>
+      <% end %>        
       <%= render Arclight::IndexMetadataFieldComponent.with_collection(presenter.field_presenters.select { |field| !compact? || field.field_config.compact }) %>
     </div>
   </div>
--- a/app/components/arclight/search_result_title_component.html.erb
+++ b/app/components/arclight/search_result_title_component.html.erb
@@ -1,7 +1,9 @@
 <header class="documentHeader row" data-document-id="<%= @document.id %>">
   <h3 class="index_title document-title-heading col">
     <%= helpers.link_to_document @document, counter: @counter %>
-    <%= tag.span @document.extent, class: 'al-document-extent badge' if @document.extent && !compact? %>
+    <% @document.extent.each do |extent| %>
+      <%= tag.span extent, class: 'al-document-extent badge' unless compact? %>
+    <% end %>    
     <%= tag.span class: 'al-document-container text-muted ml-auto' do %>
       <%= @document.containers.join(', ') %>
     <% end if @document.containers.present? %>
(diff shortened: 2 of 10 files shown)


seanaery:
@marlo-longley This is great progress and it moves this issue closer to fixed. A few thoughts...

### Multiple extents
Here's what I'd expect the result to become, i.e., rather than a single-item array, the doc should have multiple extents and only the first should concatenate the values from within the same `<physdesc>`:
```
expect(result['extent_ssm']).to equal_array_ignoring_whitespace ['1.25 Linear Feet (1 volume)', '1 document case', '16 DVDRs']
```
We have a related issue to tackle in #997 where we should be showing a badge for each extent, i.e., multiple extents should be OK as long as we fix that other issue.

### Removing `and`
With multiple values the `and` is coming from the default `separator_options` but can be overridden like has been done for several other fields in the `catalog_controller.rb` template. I.e.,
```
separator_options: {
      words_connector: '<br/>',
      two_words_connector: '<br/>',
      last_word_connector: '<br/>'
    }
```

### Repeating for `extent_tesim`
In the traject config you can just copy the value that was captured for `extent_ssm` into `extent_tesim` without repeating the logic. I.e.,
```
to_field "extent_tesim" do |_record, accumulator, context|
  accumulator.concat context.output_hash["extent_ssm"] || []
end
``` 
### Needed on components and collections alike
Since `<extent>` can be present in the top-level collection and in individual components, we'll need this tweak to also be present in `ead2_component_config.rb`; See: https://github.com/projectblacklight/arclight/blob/main/lib/arclight/traject/ead2_component_config.rb#L158-L159

seanaery:
I definitely welcome additional feedback from @mmmmcode , @gwiedeman and anyone else who can advise on the EAD nuances around extents in #989 :-)

marlo-longley:
Awesome thanks so much @seanaery !

marlo-longley:
HI @seanaery  -- I believe I addressed all your comments. I also added code to close https://github.com/projectblacklight/arclight/issues/997, so now this PR addresses 2 issues. I would love feedback but here is what it's looking like: 
Screenshot 2023-12-07 at 4 56 08 PM
Screenshot 2023-12-07 at 4 39 00 PM

The one question I had was about the failing test. I don't really understand this what this test is trying to achieve by using the `extent` field:
https://github.com/projectblacklight/arclight/blob/0d73b631a8be6f864de6df19aaaab091a73c88ae/spec/models/arclight/document_downloads_spec.rb#L13

I am happy to edit the test to make it pass, but I am getting the feeling this was potentially a workaround that we can get rid of altogether?

mmmmcode:
@seanaery @marlo-longley One thing I wonder about is the parentheses in the container summary text (<extent altrender="carrier">)- are those something we want to supply if they aren't present? Or is that something that should be addressed with local customizations?

marlo-longley:
@mmmmcode Thanks for taking a look! In these fixtures the words in parentheses are provided directly from the XML, as if a cataloger had entered it that way. So I wasn't thinking we should alter that. I want to make sure I'm understanding you, though!

mmmmcode:
From my perspective, the () provided by the cataloger looks like style
baked into content (like using <emph> to bold series titles), which I'm
always a fan of eliminating. But that might just be me ;)

marlo-longley:
Thanks for explaining @mmmmcode -- I could see this for sure but I'm thinking it is a change that is not directly related to allowing for multiple extents, or properly grouping extents, as the tickets lay out. If we want to change the formatting of text in extents, my take is thag we should open a new ticket and discuss there, rather than lump new functionality into this particular PR.

seanaery (approved):

Issue #1430, opened 2023-12-04
Grouped search results header has invalid markup, needs dl wrapper


seanaery:
In the [index_metadata_field_component.html.erb](https://github.com/projectblacklight/arclight/blob/main/app/components/arclight/index_metadata_field_component.html.erb) view there are `<dt>`/`<dd>` pairs getting rendered without a wrapping `<dl>` element. This is invalid HTML and is getting flagged as an accessibility issue by axe DevTools.

<img width="1142" alt="Screenshot 2023-12-04 at 10 49 09 AM" src="https://github.com/projectblacklight/arclight/assets/3933756/9095427c-dbd5-4e14-a426-3ee5a4da9fb0">

More:
https://dequeuniversity.com/rules/axe/4.8/dlitem?application=AxeChrome

lfarrell:
Fix PR: https://github.com/projectblacklight/arclight/pull/1470

Issue #1395, opened 2023-01-24
Invalid HTML structure in search result item


ggeisler:
Within the search result item structure there is this code for displaying the breadcrumb links/path to the item:

```html
<dl class="document-metadata dl-invert row">
    <div class="breadcrumb-links">
         <ol class="breadcrumb">
           <li class="breadcrumb-item">
              ...
```

This raises an accessibility error, and is also just invalid HTML, I believe. The `dl` can only contain `dt` and `dd` elements, so the child `div` is invalid, as is the subsequent `ol`. 

Probably fixable just by changing the `dl` to a `div`? However, if the result item has metadata to display aside from the breadcrumb path (e.g., an abstract/scope value), that part is also within the `dl` and includes an additional invalid `div`:

```html
<dl class="document-metadata dl-invert row">
    <div class="row">
       <dt class="visually-hidden">Abstract Or Scope</dt>
         <dd class="col al-document-abstract-or-scope truncator truncated" data-controller="arclight-truncate">
            <div class="content" data-arclight-truncate-target="content"><p>"Broadcasting", mainly ...
           ...
```

Basically it seems we're using a `dl` for the result item metadata following the basic Blacklight pattern (where the `dl` is used correctly and appropriately for displaying several field labels and values), but ArcLight result items display very different item metadata, so the `dl` ends up interspersed with a lot of invalid elements.

Perhaps the whole result item structure in ArcLight should be rewritten (ignoring the Blacklight way of doing it) with an eye on valid HTML.

The code in question is what renders this:

<img width="852" alt="Screen Shot 2023-01-24 at 11 01 24 AM" src="https://user-images.githubusercontent.com/101482/214372261-fcabad1f-7a03-4af4-8f81-440afd9772e7.png">

cbeer:
![Screen Shot 2023-01-30 at 18 30 40](https://user-images.githubusercontent.com/111218/215646882-78bc0a3f-eca5-4b13-9db4-8763e7a0e65a.png)

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

🤔

ggeisler:
Ah, interesting. One of the accessibility checking tools I used (Axe DevTools) says this:

> dl element has direct children that are not allowed: div > ol

Either way, based on the documentation, yes, I think the second part of my initial description that says we shouldn't have a `div` within the `dl` is incorrect.

[...]

Issue #1393, opened 2023-01-20
Unlink currently selected item in collection contents (content inventory) tree


ggeisler:
The collection inventory tree in the Collection contents section of the collection sidebar has an accessibility issue that I think we can solve and improve the user experience around it at the same time. We currently indicate which item in the collection hierarchy is currently selected by adding a yellow highlight background:

<img width="402" alt="Screen Shot 2023-01-20 at 4 08 10 PM" src="https://user-images.githubusercontent.com/101482/213821529-0bcdfd7a-7ed6-4a10-bfee-e35727cc87e7.png">

However, the default Bootstrap blue link color does not have sufficient contrast against the yellow background to meet accessibility standards. 

At the same time, I think it's deceiving, and potentially confusing, to link the selected item. It's already selected, so we shouldn't suggest to the user that they might want to select it again (doing so just refreshes the page to the same state).

I think we can solve both of those issues by removing the link from the currently selected component. If it's not a link, it 1) won't have the unnecessary link styling and 2) the text will be displayed in the default body text color, which has sufficient contrast against the yellow highlight background to pass the accessibility standard.

After removing the link from the currently selected component, the tree should look like this:

<img width="402" alt="Screen Shot 2023-01-20 at 4 07 30 PM" src="https://user-images.githubusercontent.com/101482/213822196-842698fc-8d63-4e58-aed7-e3d4d9a30424.png">

lfarrell:
Fix PR: https://github.com/projectblacklight/arclight/pull/1469

PR #1447, opened 2023-12-07
Accessibility issues

lfarrell:
Fix for: https://github.com/projectblacklight/arclight/issues/1395
Fix for https://github.com/projectblacklight/arclight/issues/1393
Fix for https://github.com/projectblacklight/arclight/issues/1430

Also addresses color contrast issues and duplicate ids in search results

Should fix all accessibility issues in arclight.

### DIFF
--- a/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
@@ -22,12 +22,6 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
   }
 }
 
-.al-hierarchy-highlight > .documentHeader {
-  a {
-    font-weight: bold;
-  }
-}
-
 #collection-context {
   ul {
     list-style: none;
--- a/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
@@ -22,6 +22,10 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
   }
 }
 
+.al-hierarchy-highlight > .documentHeader {
+  font-weight: bold;
+}
+
 #collection-context {
   ul {
     list-style: none;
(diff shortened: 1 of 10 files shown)


app/components/arclight/search_result_breadcrumbs_component.html.erb
```diff
@@ -1,3 +1,7 @@
-<div class='breadcrumb-links'>
-  <%= rendered_breadcrumbs %>
-</div>
+<dt>
```
seanaery, line 1: I like where you're going with it, but as far as I understand it, a `<dd>` can't be a child of `<dt>`.  The missing indentation makes it a bit confusing. I wonder if a small tweak like this would work:

```
<dt class="visually-hidden">Collection Context</dt>
<dd>...</dd>
```

lfarrell: @seanaery That seems to work. I've pushed up the change.

app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
```diff
@@ -22,10 +22,6 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
   }
 }
 
-.al-hierarchy-highlight > .documentHeader {
```
seanaery, line 25: I would leave this in -- I think it is useful to visually highlight the currently selected item. And now that you have unlinked the current highlighted item, its text color is black so there's no contrast issue to resolve on the highlighted document.

seanaery:
This is looking great and I am excited to get all these fixes into core. These are big important wins -- thanks for working through all of this.

I completed a review and left a few inline comments. Just a couple more thoughts, taking an updated look at the branch:
1. I believe we should keep the background-color highlight on the currently-viewed item in the sidebar; I think it's especially helpful if you have a collection with a deep hierarchy and you're trying to navigate between different component nodes at different levels. 
2. There's a lot here in one PR with 10+ commits, some of which undo or supplement the effects of others. When this gets merged it'd be a shame to squash this all into one big "Accessibility issues" commit, but leaving it un-squashed would also leave a bit of a confusing paper trail.

I wonder if you might either 1) break the changes from this PR out into 5-7 smaller separate PRs; or alternatively 2) make a copy of the branch, do an interactive rebase, e.g. `$ git rebase -i HEAD~15` and edit down / reorder the commit history with `fixup`, `pick`, `reword`, etc. so each of the final fixes here has a clear paper trail, and then post that branch for one new PR.

lfarrell:
This PR has been split out into multiple PRs, which replace this PR.

- https://github.com/projectblacklight/arclight/pull/1468
- https://github.com/projectblacklight/arclight/pull/1469
- https://github.com/projectblacklight/arclight/pull/1470
- https://github.com/projectblacklight/arclight/pull/1471
- https://github.com/projectblacklight/arclight/pull/1472
- https://github.com/projectblacklight/arclight/pull/1473

Issue #1429, opened 2023-12-01
Bookmark checkboxes don't correctly reflect whether an item is bookmarked


seanaery:
In Arclight 1.0.1 if you use the bookmark checkboxes to bookmark collections or components, the actual bookmarking succeeds, but after a refresh or subsequent pageview, wherever the bookmarked item appears, the box is no longer checked.

Screenshots demonstrating the bug...

### 1. Search
Search for `test`:
<img width="1370" alt="Screenshot 2023-12-01 at 2 25 09 PM" src="https://github.com/projectblacklight/arclight/assets/3933756/0f79a4d6-a590-4fa7-a900-3ce08c6eef62">

### 2. Check Items
Check the bookmark checkboxes (they get added to Bookmarks):
<img width="1385" alt="Screenshot 2023-12-01 at 2 25 41 PM" src="https://github.com/projectblacklight/arclight/assets/3933756/ea519ba2-e646-40ac-a568-7942ed1d0d69">

### 3. Refresh
Checkbox state is incorrect
<img width="1378" alt="Screenshot 2023-12-01 at 2 25 50 PM" src="https://github.com/projectblacklight/arclight/assets/3933756/6b97853c-83f8-41c9-b278-92de16ab1211">

### 4. View Bookmarks
The checkbox state is even incorrect on the Bookmarks page, which makes it challenging to uncheck/remove bookmarks.
<img width="1366" alt="Screenshot 2023-12-01 at 2 33 50 PM" src="https://github.com/projectblacklight/arclight/assets/3933756/44b020d1-4b81-4102-9d10-fea687ab8b71">

lfarrell:
Issue resolved by https://github.com/projectblacklight/arclight/pull/1465

PR #1465, opened 2023-12-08
Fix issue with bookmarked records not staying checked

lfarrell:
Fix for: https://github.com/projectblacklight/arclight/issues/1429


marlo-longley:
Hi @lfarrell -- thanks for this fix. It seems right to me especially based on the history in the BL repo. I wonder if we should add a similar test?

https://github.com/projectblacklight/blacklight/pull/2925/files

marlo-longley (approved):

Issue #1460, opened 2023-12-08
Display <otherfindaid> at the component level


marlo-longley:
Archivists have confirmed that this is not necessary to index into full text search at the component level. So this ticket is for display only.

This ticket is broken out of #898

marlo-longley:
Closed by https://github.com/projectblacklight/arclight/pull/1466/

Issue #1459, opened 2023-12-08
Display fileplan at collection and component levels


marlo-longley:
[fileplan](https://www.loc.gov/ead/tglib/elements/fileplan.html)

- [x] Display at the collection level
- [x] Display at the component level

This ticket is broken out of https://github.com/projectblacklight/arclight/issues/898

marlo-longley:
Closed by https://github.com/projectblacklight/arclight/pull/1466

Issue #1456, opened 2023-12-08
Display <odd> 


marlo-longley:
This is a widely used field.

This ticket is broken out of https://github.com/projectblacklight/arclight/issues/898

marlo-longley:
In searching the codebase I am not finding any `<odd>` in our fixture data. I think this needs further confirmation that it is being displayed.

gwiedeman:
There is an `<odd>` here: https://github.com/projectblacklight/arclight/blob/91420125acc80381502a8b0996b39b931f3bf4aa/spec/fixtures/ead/nlm/alphaomegaalpha.xml#L77

But I didn't add tests for these, which could be useful.

marlo-longley:
Thank you!! Oops.

Issue #1452, opened 2023-12-08
Handle bioghist at collection and component levels


marlo-longley:
[bioghist](https://www.loc.gov/ead/tglib/elements/bioghist.html)

We are already indexing this field into the searchable text fields bucket on the collection level.

Tasks

 - [x] Display at the collection level
 - [x] Index at the component level
 - [x] Display at the component level

This ticket is broken out of https://github.com/projectblacklight/arclight/issues/898

gwiedeman:
Will soon display in component level and added example to fixtures: http://localhost:3000/catalog/aoa271aspace_563a320bb37d24a9e1e6f7bf95b52671

marlo-longley:
Confirmed this is not displaying at the collection and component level, closed by https://github.com/projectblacklight/arclight/pull/1466/files

PR #1466, opened 2023-12-08
Adds missing elements

gwiedeman:
This adds the straightforward missing elements, mostly at the component level.

This should address #1450, #1451, #1452, #1456, #1459, and #1460. Also added some test examples for many.


seanaery:
Thanks for knocking these out! One quick thing -- I think we will need to add labels for these new fields in https://github.com/projectblacklight/arclight/blob/main/lib/generators/arclight/templates/config/locales/arclight.en.yml else a field like `odd` will probably render with a header like `Odd` instead of something intuitive like `Other descriptive data`.

seanaery (approved):
Excellent -- and thanks for supplementing the fixtures!

PR #1471, opened 2023-12-11
Fix issue with duplicate ids on the page

lfarrell:
Split out from: https://github.com/projectblacklight/arclight/pull/1447

### DIFF
--- a/app/views/catalog/_document_list.html.erb
+++ b/app/views/catalog/_document_list.html.erb
@@ -2,7 +2,7 @@
 <% # https://github.com/projectblacklight/blacklight/blob/v8.1.0/app/views/catalog/_document_list.html.erb %>
 <% # container for all documents in index list view -%>
 <% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
-<div id="<%= 'documents' unless grouped? %>" class="al-document-listings documents-<%= view_config&.key || document_index_view_type %>">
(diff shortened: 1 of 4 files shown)


app/views/catalog/_document_list.html.erb
```diff
@@ -0,0 +1,8 @@
+<% # Override to remove id="documents" on div. When documents are grouped it causes duplicate id issues %>
+<% # https://github.com/projectblacklight/blacklight/blob/v8.1.0/app/views/catalog/_document_list.html.erb %>
+<% # container for all documents in index list view -%>
+<% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
+<div class="al-document-listings documents-<%= view_config&.key || document_index_view_type %>">
```
seanaery, line 5: After looking closer, I think we'll need to keep the `documents` id in a non-grouped view of search results (to accommodate the `Skip to First Result` skip-link in those views). Here's a recommended way to only remove it only when the results are grouped...
```
<div id="<%= 'documents' unless grouped? %>" class="...">
```

seanaery (changes requested):
Just one change to make, per my inline comment/suggestion. That `Skip to First Result` link will have to work whether in grouped or ungrouped view...

app/views/catalog/_document_list.html.erb
lfarrell: Updated. I modified it slightly `<div <%= 'id="documents"'.html_safe unless grouped? %>` to avoid having empty ids for grouped records, `id=""`

seanaery (approved):

marlo-longley (approved):

PR #1472, opened 2023-12-11
Show "View More/View Less" links as a darker color

lfarrell:
Fixes contrast issue

Split out from: https://github.com/projectblacklight/arclight/pull/1447

### DIFF
--- a/app/assets/stylesheets/arclight/modules/truncator.scss
+++ b/app/assets/stylesheets/arclight/modules/truncator.scss
@@ -28,12 +28,12 @@
 
   // "view more"/"view less" text swaps when content is expanded
   .view-more {
-    color: $blue-600;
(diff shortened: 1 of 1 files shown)


seanaery (changes requested):
Minor quibble here, but I would generalize this more so institutions who are not using the default Bootstrap blue for links have a better chance of not having to override this style locally.

Looking at [Bootstrap's variables](https://github.com/twbs/bootstrap-rubygem/blob/main/assets/stylesheets/bootstrap/_variables.scss#L84C12-L84C35) it looks like `$blue-600` is equivalent to shading the default link color 20% darker, so this should work well: 
```
color: shade-color($link-color, 20%);
```

seanaery (approved):

marlo-longley (approved):

Issue #532, opened 2019-07-22
Analyze and improve relevance ranking


anarchivist:
Users regularly report that they don't know why certain results are being shown first 

* [x] Document factors leading to search relevance #597 
* [ ] Relevance: adjust boost on titles for components/collections #723
* [ ] Relevance: provide documentation on Solr configuration (especially boost/query analysis) #724
* [x] Relevance: analyze stemming behavior #725
* [x] Better support known-item searching for collections #726
* [ ] Relevance behavior grouped results #727

mejackreed:
Related to #531

seanaery:
Closing out this ticket as we have implemented several relevance improvements in the [Dec 2023 community sprint](https://github.com/orgs/projectblacklight/projects/3/views/2). We have also updated the wiki doc here: https://github.com/projectblacklight/arclight/wiki/Configuring-Solr-and-Search-Relevancy

PR #1476, opened 2023-12-11
Improve and document use of mm (minimum should match) param in Solr c…

seanaery:
…onfig. Advances #532.

Fixes these problems with the previous config:
 - only 2 terms had to match in a 3-term query, or 3 terms out of 4 (too permissive)
 - any 3-10 term query had to match all but 1 term *except* for 6-term queries, which oddly had to match all but 2. Now all terms must match for a query of 4 or fewer. Above 4, 90% must match (rounded down).

### DIFF
--- a/solr/conf/solrconfig.xml
+++ b/solr/conf/solrconfig.xml
@@ -75,7 +75,15 @@
        <int name="rows">10</int>
 
        <str name="q.alt">*:*</str>
-       <str name="mm">2&lt;-1 5&lt;-2 6&lt;90%</str>
+
+       <!-- Minimum Should Match parameter (mm) -->
+       <str name="mm">4&lt;90%</str>
+       <!-- 1-4 term query, all must match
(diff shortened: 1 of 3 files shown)


randalldfloyd (approved):
This desk-checks for me in the UI.  

I tested by:

- Indexing alphaomegaalpha.xml from spec fixtures
- Obtained results with "alpha omega oath" 
- Obtained NO results with "alpha omega oath nonsense"
- Obtained results again with "alpha omega oath fraternity nonsense"

marlo-longley:
I ran the same process and got the same results. I can approve this code-wise, but not sure if any implementers have thoughts or feelings about the change.

marlo-longley (approved):

randalldfloyd:
@seanaery In addition to my approval based on observance, I would also add that the new behavior feels right to me from a user perspective.  I would vote for accepting/merging if no additional reviewers want to ring in on it.

PR #1477, opened 2023-12-11
Set the default SearchHandler to just return all fields. Advances #532

seanaery:
- this is the general intent in arclight, see * in config.default_solr_params & config.catalog_controller_group_query_params
- simplifies managing this config as fields change over time.


lfarrell:
This does seem better than having to remember to add any new/custom fields

lfarrell (approved):

marlo-longley (approved):

Issue #1028, opened 2021-03-01
Single dates exported from ArchivesSpace do not display


gwiedeman:
By default ArchivesSpace permits three date types, inclusive, bulk, and single (and you can add more local types). Only inclusive and bulk are valid in EAD2002, so single dates export to EAD without a @ type.

During indexing, [this method](https://github.com/projectblacklight/arclight/blob/master/lib/arclight/normalized_date.rb#L41-L49) parses unitdates and adds them to normalized_date which is used for display.

Currently, if a component has an inclusive date and a single date, only the inclusive date will be added to normalized_date and the single date will not display to users.

seanaery:
Thanks for reporting, @gwiedeman . We noticed this issue, too, at Duke and made local modifications (esp. to the `normalize` method):
https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/blob/develop/lib/dul_arclight/normalized_date.rb
https://gitlab.oit.duke.edu/dul-its/dul-arclight/-/blob/develop/spec/lib/dul_arclight/normalized_date_spec.rb

Seems to be working well for us so far, but I imagine there might be variations throughout the community on how dates are treated.

gwiedeman:
relates to #1336

gwiedeman:
I have this up in [single_dates](https://github.com/projectblacklight/arclight/tree/single_dates), and @seanaery 's fix ensures that all `<unitdate>`s display, even when some have `@type` and some do not. However, this doesn't always preserve the order of the dates, as some archivists would expect. For example, [the fixture I added](https://github.com/projectblacklight/arclight/blob/single_dates/spec/fixtures/ead/nlm/alphaomegaalpha.xml#L412-L413) has a `<unitdate>` without a `@type` and then an undated `<unitdate>` with `@type="inclusive"` Since Arclight currently indexes dates into individual fields by `@type` this just appends all the inclusive dates, then dates without `@type` and bulk dates are appended last. If all your dates have the same `@type` or no `@type` the order will be preserved, but not if they're mixed.

Changing this to always preserve date order would require reworking how Arclight handles dates. It is debatable how important the order of unitdates are. They do get exported from ASpace following the same order in the webapp, but it's unclear if EAD is promising to preserve date order. Many archivists do think date order is meaningful and enter dates how they expect is most intuitive for users. Many would expect that date order is maintained.

[...]

PR #1486, opened 2023-12-13
All <unitdates> now display

gwiedeman:
Implemented Duke's fix so that all `<unitdates>` are displayed, even when only some have `@type`. Also added example fixture.

This does not necessarily preserve `<unitdate>` order unless they all have the same `@type`, so #1028 should remain open.


seanaery (approved):

randalldfloyd (approved):
This checks out for me locally.

PR #1487, opened 2023-12-14
Improve search relevance by enabling queries for ids, containers, lan…

### LINKED ISSUE #894: Verify which fields to search via "All Fields" search
As @cmkpowell noted today, text in the `<langmaterial />` element (indexed as `language_ssm` & `language_sim` at top-level, just `language_ssm` on components) does not currently get searched when performing an `All Fields` search.

**Use Case:** 

- [This collection](https://arclight-demo.projectblacklight.org/catalog/umich-scl-rosenberg) has the word `Yiddish` in its top-level `<langmaterial />` element. However, the collection is not found when [searching all fields for `yiddish`](https://arclight-demo.projectblacklight.org/catalog?group=true&q=yiddish&search_field=all_fields).

Current list of fields included in an All Fields search can be found in `schema.xml`:
https://github.com/projectblacklight/arclight/blob/master/solr/conf/schema.xml#L338-L370

This could be reconciled against fields that are available (see `[traject/ead2_config.rb](https://github.com/projectblacklight/arclight/blob/master/lib/arclight/traject/ead2_config.rb)`) to identify any other missing but desired fields.

seanaery:
…guage, & ancestor component titles. Closes #894 & Advances #532.

- Adds Container & Identifier to search scope dropdown
- Adds parent component titles, language, containers, & identifiers into the default all-fields search
- TODO: more scoring boost refinement; this PR only gets these elements to be considered at all in queries

### DIFF
--- a/lib/generators/arclight/templates/catalog_controller.rb
+++ b/lib/generators/arclight/templates/catalog_controller.rb
@@ -231,6 +231,20 @@ class CatalogController < ApplicationController
         pf:  '${pf_title}'
       }
     end
+    config.add_search_field 'container', label: 'Container' do |field|
+      field.qt = 'search'
+      field.solr_parameters = {
+        qf:  '${qf_container}',
+        pf:  '${pf_container}'
+      }
+    end
+    config.add_search_field 'identifier', label: 'Identifier' do |field|
+      field.qt = 'search'
+      field.solr_parameters = {
+        qf:  '${qf_identifier}',
+        pf:  '${pf_identifier}'
+      }
+    end
 
     # These are the parameters passed through in search_state.params_for_search
     config.search_state_fields += %i[id group hierarchy_context original_document]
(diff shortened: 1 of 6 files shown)


seanaery:
After an analysis of what fields were being included vs. omitted in All Fields searching, the vast majority of fields were indeed accounted for. However, these stood out as glaring omissions:
- containers
- language
- parent titles (i.e., the titles from a component's ancestor breadcrumb trail)
- ids (including EAD ID, component ref ID, the concatenated ID that ends up in the URL, aka `id` ) 

This PR adds these fields for discovery _without making any assertions about relative relevance boosts_, as decisions there may take some time to sort out.

The six query specs included here in `fielded_search_results_spec.rb` all would have returned **no results** prior to this PR.

lfarrell (approved):

seanaery:
There's a test in here that uses the non-underscore component URL. I'll update it shortly

marlo-longley (approved):

PR #1491, opened 2023-12-14
Display Related section at component level

### LINKED ISSUE #1464: Display <relatedmaterial> <originalsloc> and <separatedmaterial> at component level
[relatedmaterial](https://www.loc.gov/ead/tglib/elements/relatedmaterial.html)
[separatedmaterial](https://www.loc.gov/ead/tglib/elements/separatedmaterial.html)
[originalsloc](https://www.loc.gov/ead/tglib/elements/originalsloc.html)

We have them indexed but not displayed.

Broken out from #898

marlo-longley:
Closes #1464

To be honest, I don't understand why it was necessary to add a whole new ViewComponent to achieve this. I had to follow the exact pattern that already existed for `AccessComponent` to get this to display as its own section at the (EAD) component level. Initially I was able display it as part of the upper metadata with much less code, but I wanted it to look similar to the top collection level display, which puts "Related" as its own section. 

Open to feedback on doing this in a simpler way! If this seems like the right approach, I can add some tests as well. 

Here is the result of this in the UI 
Screenshot 2023-12-14 at 3 44 56 PM

### DIFF
--- a/app/components/arclight/document_component.html.erb
+++ b/app/components/arclight/document_component.html.erb
@@ -36,4 +36,6 @@
 </div>
 <% end %>
 
+<%= related %>
+
 <%= access %>
--- a/app/components/arclight/document_component.rb
+++ b/app/components/arclight/document_component.rb
@@ -38,6 +38,10 @@ module Arclight
       render (blacklight_config.show.access_component || Arclight::AccessComponent).new(presenter: presenter)
     end
 
+    def related
+      render (blacklight_config.show.related || Arclight::RelatedComponent).new(presenter: presenter)
+    end
+
     def toggle_sidebar
       button_tag(t('arclight.views.show.toggle_sidebar'),
                  type: :button,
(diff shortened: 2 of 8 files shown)


seanaery:
@marlo-longley That is an interesting approach and it certainly works. It does add some complexity that is probably not needed. The approach in arclight to date for component metadata has typically been to just put it all in one common bucket for display rather than try to segment it out into grouped sections.

This differs from the segmented collection metadata approach, but for good reason. Collections in EAD are often quite loaded with metadata, and the collection nav includes useful links to get to the different sections.

By contrast, components in EAD are in practice often only sparsely described. Though EAD _permits you to_ have a lot of description on the component level, from what I have seen, the most common thing is a component will likely only have 1-3 total fields that'd get rendered in its metadata section.

marlo-longley:
@seanaery I would love not to add this to the repo haha. Thank you. I guess my assumption that it needed to look like the collection-level UI was incorrect.

marlo-longley:
@seanaery  -- I will just add these as `component_field` as I think you are suggesting. I'll close this and open a new PR.

seanaery:
Sounds great -- thank you! Yep, I think we can simply add these right alongside all the existing `config.add_component_field` fields.

PR #1492, opened 2023-12-14
Draft: Don't load all hierarchical items at once when expanding the tree

lfarrell:
This should work to speed up loading of large sibling lists. It essentially paginates the results. It bumps up the results size from 10 to 100 and adds a button for every subsequent 100 items with a truncated document title as part of the button display for navigation context. This can result in a LOT of buttons for REALLY long lists

Screen Shot 2023-12-14 at 3 41 37 PM

I'm not sure it's a great idea to do it this way, but if folks think it's worth pursuing. I can continue down this path.

### DIFF
--- a/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
+++ b/app/assets/stylesheets/arclight/modules/hierarchy_and_online_contents.scss
@@ -26,6 +26,15 @@ $hierarchy-view-collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3
   background: $mark-bg;
 }
 
+li.collapse .al-hierarchy-more {
+  display: none;
+}
+
+li.collapse.show .al-hierarchy-more {
+  display: block;
+  margin-bottom: 5px;
+}
+
 #collection-context {
   ul {
     list-style: none;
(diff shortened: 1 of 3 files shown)


gwiedeman:
I showed this to a colleagues who was skeptical. They thought that if the archivist wanted to break it up, they would have divided it up into more series.

1k components definitely get slow and this is a good thought, but it might not be intuitive to a user what "Items 301 to 400" means. Is it possible to display the first 100 results and have the "expand" or "next" button append the next 100 continually instead of loading it all? Then the user could keep clicking "expand."

lfarrell:
@gwiedeman @seanaery I've updated it to only show one expand button at a time. It would have been more straightforward in arclight 0.x, but with the move to turbo-frames in 1.x it's a little trickier.
This branch would probably need to be cleaned up a bit to move the JS out of the partial.

seanaery:
@lfarrell I do like the updated one-button-at-a-time UI better, and I agree that we ultimately wouldn't want the JS in the partial itself. Trying this branch out a bit with real finding aids, I also notice a few cases that are not yet accounted for here:
- If the 1,000s of sibling components are at the very top c01 level, this code doesn't truncate them. When you view the collection page and look at the sidebar, that will spin for a long time and either timeout or eventually render all the components all at once if it can complete before the timeout limit.
- If you view a nested component page, say, one that's number 300 in a sequence of 1,000 or more siblings, the bottom Expand button in the sidebar doesn't limit the number it will load, so clicking it will try to just load the rest, which will either timeout or take a long time.

Issue #1490, opened 2023-12-14
Is the <acqinfo> (acquisition information) field behaving as intended?


seanaery:
This ticket was spawned out of #784 as it is still in need of verification.

### Current Behavior

The **Acquisition Information** field (`<acqinfo>` in EAD) currently behaves like none other in arclight. Is this intentional or should we change this? 
- When a component is being indexed, it always gets its owning collection's `<acqinfo>` value. It also gets it own `<acqinfo>` values (if present). [See traject rules here](https://github.com/projectblacklight/arclight/blob/main/lib/arclight/traject/ead2_component_config.rb#L272-L275).
- A component does not inherit any of its ancestor components' `<acqinfo>` values, only its top-level collection's `<acqinfo>`.
- This means on every single component page, the metadata includes `Acquisition information`. [Demo Example](https://arclight-demo.projectblacklight.org/catalog/umich-bhl-0312aspace_66733ea1d7a9837dfba9e48d32c4d61b). 
- There are no other fields that behave this way, where the component's value is inherited from the value at its collection level.

gwiedeman:
In practice, archival systems have not inherited description like this. Instead the user is expected to infer that the upper level description applies. In a paper finding aid, a user would be expected to assume the collection level acquisition statement in the front matter applied to everything basically. Depending on the design of the system, this expectation may or may not be reasonable.

However, [DACS](https://saa-ts-dacs.github.io/dacs/06_part_I/02_chapter_01.html#requirements-for-multilevel-descriptions) _does_ say that every element must have [every required multilevel element](https://saa-ts-dacs.github.io/dacs/06_part_I/02_chapter_01.html#multilevel-required), but if its the same as the level above it should be omitted and the implicit inheritance applies. I think many archivists are not totally aware of this, and its kind of treated as a technicality.

With Arclight, we can probably manage this implicit inheritance better than archivists have in the past, as its not always clear to users. We're already doing that for access and use restrictions, it just displays in a separate section. This is a great example of better leveraging inheritance, as users shouldn't be expected to click to another page to see if the thing they are using is restricted.

[...]

PR #1496, opened 2023-12-15
Do not have components inherit collection-level acqinfo values. Close…

seanaery:
…s #1490


marlo-longley (approved):

lfarrell:
Nice! We've hit this issue locally. It'll be good to undo our customization.

PR #1505, opened 2023-12-18
Adding bibliography/bibref elements to indexers and catalog views

### LINKED ISSUE #1458: Display bibliography at collection and component levels
- [ ] Display at the collection level
 - [ ] Display at the component level

This ticket is broken out of https://github.com/projectblacklight/arclight/issues/898

randalldfloyd:
Closes #1458 

Adding missing bibliography elements from collection and component indexers, and also to catalog views.

Per feedback on #1458, this solution skips over potentially elaborate bibliography text and just creates lists from child `bibref` elements instead.


seanaery (changes requested):
This is close but it looks like it is not rendering the italic formatting despite using the `render_html_tags` helper. A couple recommended revisions to that end:
- All the fields that we're indexing where we're retaining the EAD tags for transformation to HTML should have a `_html_tesm` suffix, so this should be `bibref_html_tesm`. To keep the tags in there, the traject rule also needs `to_text: false` (else it will strip all tags and just keep the text).

Issue #1481, opened 2023-12-12
Indexing Docs/Wiki should be updated


gwiedeman:
It looks to me like the [Indexing EAD in ArcLight](https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight) wiki was mostly written before the Traject workflow. It does document how to [index with Traject](https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight#using-traject), but its a bit hidden on the bottom. It now seems to still prioritize `rake arclight:index` which I don't think works anymore, although if it does that should be a bit clearer. This could clear up some confusion for new users.

Additionally, the Traject indexing commands both on [the wiki](https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight#using-traject) and [the README.md](https://github.com/projectblacklight/arclight#traject-indexing-of-ead-content), assume the Traject config file is in `lib/arclight/traject/ead2_config.rb`. For a lot of setups, that command would raise an error. For example, if you installed a new app with the template, I believe the config file gets stored wherever your gems live. For installs with `rvm` that might be in `~/.rvm/gems/ruby-3.2.2@gemset-name/gems/arclight-1.0.1/lib/arclight/traject/ead2_config.rb` or similar. If would be cool if there was an easier way to manage this in practice to recommend (maybe a symlink?), but if there's no good ideas here that don't cause other issues, just explaining where the config might be via `echo $GEM_HOME` or similar in the docs might avoid some headaches for new users.

PR #1510, opened 2023-12-19
Updated indexing in README

gwiedeman:
Updated the README to simplify the quickstart index command. It now suggests using `rails arclight:index` as the least error-prone way of getting an EAD in the index and then links to the wiki page that has more details on how Arclight uses Traject.


gwiedeman:
Closes #1481

marlo-longley (commented):
Is it worth mentioning `index_dir` here, or that it's possible to index groups/folders of EAD in one go? We do mention it in the wiki: https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight#adding-more-finding-aids-and-repositories

PR #1515, opened 2024-03-12
Run rubocop autocorrect

corylown:
New version of rubocop wants some changes. This will get CI working again.


dl-maura (approved):
Ran locally correctly!

PR #1531, opened 2024-05-04
Use collection_id for linking rather than inferring from eadid

corylown:
At the last community call (2024-04-22) we discussion a desired feature to make it easier to customize the formation of the collection document id. The use cases cited were needing to add the repository name/code to prevent identifier collisions across repositories and adding a slug version of the title to the id for friendlier URLs.

As I was looking at how we could better support this I found that there were a couple places in the application where we were inferring the parent collection id from the normalized form of the eadid. I found this confusing and was concerned that allowing customization of the document id would break these assumptions about the relationship between the parent collection document id and the eadid.

This PR attempts to remove this assumption from the app so we can (in a future PR) add a seam for customizing the collection document id formation.

Instead of inferring the parent collection document id from the `eadid` this creates a `SolrDocument` accessor method `collection_id` using the `_root_` Solr field from Solr's nested document structure. This change works with either older style `legacy_id` parent field relationships or the newer `parent_ids` field. It should be completely backwards compatible with existing arclight v1 apps.

Grateful for feedback, questions, concerns.

### DIFF
--- a/app/components/arclight/collection_sidebar_component.rb
+++ b/app/components/arclight/collection_sidebar_component.rb
@@ -29,13 +29,11 @@ module Arclight
     end
 
     def document_path
-      @document_path ||= solr_document_path(normalized_eadid)
+      @document_path ||= solr_document_path(document.collection_id)
     end
 
     def section_anchor(section)
       "##{t("arclight.views.show.sections.#{section}").parameterize}"
     end
-
-    delegate :normalized_eadid, to: :document
   end
 end
--- a/app/models/arclight/parents.rb
+++ b/app/models/arclight/parents.rb
@@ -29,7 +29,7 @@ module Arclight
     # @param [SolrDocument] document
     def self.from_solr_document(document)
       ids = document.parent_ids
-      legacy_ids = document.legacy_parent_ids.map { |legacy_id| document.eadid == legacy_id ? legacy_id : "#{document.eadid}#{legacy_id}" }
+      legacy_ids = document.legacy_parent_ids.map { |legacy_id| document.collection_id == legacy_id ? legacy_id : "#{document.collection_id}#{legacy_id}" }
       labels = document.parent_labels
       eadid = document.eadid
       levels = document.parent_levels
--- a/app/models/concerns/arclight/solr_document.rb
+++ b/app/models/concerns/arclight/solr_document.rb
@@ -29,6 +29,7 @@ module Arclight
       attribute :total_component_count, :string, 'total_component_count_is'
       attribute :online_item_count, :string, 'online_item_count_is'
       attribute :last_indexed, :date, 'timestamp'
+      attribute :collection_id, :string, '_root_'
     end
 
     def repository_config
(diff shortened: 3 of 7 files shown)


seanaery:
@corylown This looks good. Thanks for thinking it through. I have tested these changes out locally and they work well.

I would recommend taking this refactor even further, and removing `Arclight::NormalizedId` and all references to it. It's redundant at this point. The app already normalizes the `id` (strips whitespace and turns `.` to `-`) [during indexing here](https://github.com/projectblacklight/arclight/blob/main/lib/arclight/traject/ead2_config.rb#L78), and as you've noted that value ends up as the `_root_` on all the collection & component docs. The `ead_ssi` field holds the _non-normalized_ version of the`eadid`, but that is sparsely used in the app: it factors into the `qf` in `solrconfig.xml` but otherwise isn't apparently displayed anywhere or used for anything that impacts the UI.

I.e., remove this file:
```
lib/arclight/normalized_id.rb
```

Remove this method from `app/models/concerns/arclight/solr_document.rb`, since nothing calls it anymore:
```
def normalized_eadid
  Arclight::NormalizedId.new(eadid).to_s
end
```
...or if we're concerned that implementers may have called `normalized_eadid` in their code, we could deprecate the method and change it so it just calls `collection_id`.

In `app/models/arclight/parents.rb` & `app/models/arclight/parent.rb` (these models exist only for the breadcrumb display), the `eadid` is modeled and available but is not actually used anywhere. So it could be entirely stripped out, or alternatively this...
```
def eadid
  Arclight::NormalizedId.new(@eadid).to_s
end
```
Could be:
```
def eadid
  @eadid
end
```
And this...
```
eadid = document.eadid
```

Could just be:
```
eadid = document.collection_id
```

Here's a [gist](https://gist.github.com/seanaery/27fbc1ed83a8c5cdfd05acbac4fa13ac), note: without rspec tests. I did test these changes out by checking out `v1.0.1`, rebuilding/reindexing, then checking out this new branch and _before reindexing_ verifying the breadcrumb & context hierarchy UI worked; then reindexed and verified again that it worked.

corylown:
@seanaery thanks for taking a close look at this and for your thorough testing. My initial work on this included many of the changes you suggest, then I backed them out for a few reasons.

1. I plan to make changes to `Arclight::NormalizedId` in a future PR to serve as the seam where collection id formation can be customized. It didn't make sense to me to remove it only to add it back again. A preview of where I am heading with this: https://github.com/projectblacklight/arclight/compare/main...collection-id-creation
2. As you note, my reason for leaving `normalized_ead` in place is because it's a public method and could be in use by an implementation for various reasons. We could deprecate it, but I don't think we should make it equivalent to collection_id. Adding a **kwargs argument to the Arclight::NormalizedId would support its use for either applying normalization to just the eadid or forming an id from the eadid and other components. I'm torn on this. Maybe this could be treated as a separate issue?
3. I had a similar concern about ripping `eadid` out of `app/models/arclight/parents.rb` & `app/models/arclight/parent.rb` that maybe implementers are using these models and expecting `eadid` to be there. It would be nice to make these classes leaner. Would you be open to considering this change in a separate PR? I'd like to keep this one simple and not make too many changes at once.

seanaery (approved):
That all sounds good to me. 👍

PR #1532, opened 2024-05-08
Supply title and repository as kwargs for local id forming requirements

corylown:
This builds on #1531 to provide a way for applications to customize how the collection id is formed, making it possible to add either the title or repository to the id to fulfill local requirements. This does not change how the default id is formed from the eadid.

Here's an example of how a local application could use this to add a slug form of the title to the collection id: https://github.com/sul-dlss/stanford-arclight/compare/add-title-to-id?expand=1

### DIFF
--- a/lib/arclight/normalized_id.rb
+++ b/lib/arclight/normalized_id.rb
@@ -5,9 +5,11 @@ require 'arclight/exceptions'
 module Arclight
   ##
   # A simple utility class to normalize identifiers
-  # to be used around the application for linking
   class NormalizedId
-    def initialize(id)
+    # Accepts unused kwargs from the ead2_config.rb id to_field directive
+    # (:title and :repository) so that applications can provide a custom
+    # id_normalizer class to traject to form the collection id from these attributes.
+    def initialize(id, **_kwargs)
       @id = id
     end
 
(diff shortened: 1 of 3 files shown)


lib/arclight/normalized_id.rb
```diff
@@ -5,9 +5,11 @@
 module Arclight
   ##
   # A simple utility class to normalize identifiers
-  # to be used around the application for linking
   class NormalizedId
-    def initialize(id)
+    # Accepts unused kwargs from the ead_config.rb id to_field directive
```
marlo-longley, line 9: `ead2_config.rb`

marlo-longley:
Looks good to me @corylown. My only suggestion is that we should maybe edit this area of the Wiki about customizing IDs: https://github.com/projectblacklight/arclight/wiki/Indexing-EAD-in-ArcLight#advanced-component_identifier_format-in-arclight

Right now the Wiki / code assumes that the place to customize component ids is the `component_identifier_format` -- it remains the case that you can customize that using some combination of `ref_id` and `root_id` (changing the separator character, etc.). But this PR allows us to change `root_id` itself, which may be actually the target of where many implementers (like us at Stanford) want to make deeper changes to how IDs appear in the URLs. So anyway, some more info in the Wiki or elsewhere could help make it clear when to use which approach for customizing IDs. It doesn't necc have to be part of this PR.

lib/arclight/normalized_id.rb
corylown: fixed. thank you!

corylown:
Thanks for taking a look @marlo-longley. Yeah, we'd want to update the wiki if/when this change is released.

seanaery (approved):

PR #1542, opened 2024-06-07
Allow changing colors of the collapse/expand

jcoyne:
Without having to change the navbar color


seanaery (approved):
Nice -- that will be really useful!

PR #1547, opened 2024-06-12
Eliminate media query in favor of a bootstrap class

jcoyne:
This makes it easier for the downstream consumers to customize (by overriding the template)


seanaery (approved):

seanaery:
Thanks for these helpful changes. Looks like this one just got merge-conflicted with the previously merged PR because they change the same file in about the same spot. I think if you rebase on `main` and then take out the whole `@media (max-width: 767px) { ... }` block it'll be good to go.

PR #1560, opened 2024-10-05
Easier override for `Arclight::BookmarkComponent`

kirkkwang:
This commit will make it easier to override the `Arclight::BookmarkComponent` where all you would have to do is add it to the `CatalogController`.  Other components should reference that and have a fallback to the original `Arclight::BookmarkComponent`

### DIFF
--- a/app/components/arclight/document_component.html.erb
+++ b/app/components/arclight/document_component.html.erb
@@ -11,7 +11,7 @@
     <%= document.normalized_title %>
   <% end %>
   <%= render 'arclight/requests', document: document %>
-  <%= render Arclight::BookmarkComponent.new document: document, action: bookmark_config %>
+  <%= render (blacklight_config.index.document_actions.arclight_bookmark_control.component || Arclight::BookmarkComponent).new document: document, action: bookmark_config %>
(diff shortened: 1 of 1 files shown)


app/components/arclight/document_component.html.erb
```diff
@@ -11,7 +11,7 @@
     <%= document.normalized_title %>
   <% end %>
   <%= render 'arclight/requests', document: document %>
-  <%= render Arclight::BookmarkComponent.new document: document, action: bookmark_config %>
+  <%= render (blacklight_config.index.document_actions.arclight_bookmark_control.component || Arclight::BookmarkComponent).new document: document, action: bookmark_config %>
```
cbeer, line 14:
```suggestion
  <%= render (presenter&.view_config&.document_actions&.arclight_bookmark_control.component || Arclight::BookmarkComponent).new document: document, action: bookmark_config %>
```

I think the presenter should have the appropriate configuration context, maybe?

kirkkwang: Hey @cbeer, I don't seem to be finding what I'm looking for using the presenter view_config

image

PR #1564, opened 2024-11-08
Permits ERB syntax in `downloads.yml` file

sjproctor:
This commit fixes the `malformed format string - %=` error that occurs when ERB syntax is used in the `downloads.yml` file.

Screenshot 2024-11-07 at 4 40 09 PM

### DIFF
--- a/app/models/arclight/document_downloads.rb
+++ b/app/models/arclight/document_downloads.rb
@@ -33,7 +33,8 @@ module Arclight
     class << self
       def config
         @config ||= begin
-          YAML.safe_load(::File.read(config_filename))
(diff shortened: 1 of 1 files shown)


sjproctor:
> Hi @sjproctor -- can you say a bit more about why this change is needed? The current `downloads.yml` pattern is to use `%{}` syntax that will evaluate accessor methods set on `SolrDocument`. What's the use case that this pattern doesn't cover?

Hi there. I am looking to use an environment variable in the yml file. The ERB tags are causing an error.

corylown:
@sjproctor I think CI was not working when you made this pull request. Can you rebase from the latest main? and if the tests pass this looks okay to me.

PR #1573, opened 2025-01-08
Add ruby 3.4 to the test matrix

jcoyne:


seanaery:
Thanks @jcoyne . Looks like the failing tests here are because ruby dropped the `csv` gem from the standard library in 3.4.0. So I think we probably just need to explicitly require it in the arclight gemspec now, i.e., `spec.add_dependency 'csv'`

jcoyne:
@seanaery we don't use the CSV gem in Arclight though, so it seems wrong to fix it here. I think https://github.com/sul-dlss/traject_plus/pull/39 is what we're waiting for.

seanaery:
@jcoyne Aha, you're right -- thanks for the heads-up about `traject_plus`.

seanaery (approved):

seanaery (changes requested):
@jcoyne Looks like the CSV issue is now clear w/the `v2.0.1` version of `traject_plus` but there's a related issue with the `mutex_m` gem now:
https://github.com/projectblacklight/arclight/actions/runs/12894248578/job/35952604069?pr=1573

jcoyne:
Yes. I believe the problem is now https://github.com/traject/traject/issues/299

seanaery (approved):

